Issue #33, opened 2013-09-12
Serialized column types


outoftime:
Should support serializing arbitrary objects to JSON and messagepack at a minimum.

PR #48, opened 2013-10-05
Serialized columns

lsimoneau:
Hey Mat. I've spiked out an idea for what serialisation could look like. This isn't ready to merge, I just wanted to check if you were ok with the general direction before fleshing it out and speccing it in detail.

I initially tried putting the logic for serialization in `cast`, but that doesn't work since there's no distinction between casts on the way into the DB and on the way out. So I arrived at this idea, which is using an intermediate object that has a serializing `to_s` for insertion, and checking for the presence of that type in the attribute reader to ensure that the original object is what gets passed back.

I think it's likely the cleanest solution, but I'm not sure where you'd prefer the new classes live. I've given it a new second-level namespace, but perhaps you'd prefer it live in a sub-module inside of `Record`?

Anyway, let me know. If you're happy with the direction I'll wrap up the specs and add in messagepack to close off #33.


pezra:
I am closing this PR due to inactivity. If anyone wants to work on this feel free to create a new PR.

PR #103, opened 2014-04-14
Add STATIC column support

ryanwilliams:
Allows you to have static columns when using a clustering key.

e.g.

    class Post
      include Cequel::Record

      key :blog_subdomain, :text
      key :id, :timeuuid, auto: true
      column :blog_title, :text, static: true
      column :title, :text
      column :body, :text
      column :created_at, :timestamp
    end

### DIFF
--- a/lib/cequel/record/schema.rb
+++ b/lib/cequel/record/schema.rb
@@ -113,7 +113,7 @@ module Cequel
 
         def column(name, type, options = {})
           super
-          table_schema.add_data_column(name, type, options[:index])
+          table_schema.add_data_column(name, type, options)
         end
 
         def list(name, type, options = {})
--- a/lib/cequel/schema/column.rb
+++ b/lib/cequel/schema/column.rb
@@ -80,6 +80,13 @@ module Cequel
       end
 
       #
+      # @return [Boolean] true if this column is static
+      #
+      def static?
+        false
+      end
+
+      #
       # @param value the value to cast
       # @return the value cast to the appropriate type for this column
       #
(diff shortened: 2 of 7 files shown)


pezra:
@ryanwilliams, i know the PR is old but if you are still interested in this (i know i am) please rebase it onto master. Once the tests pass i am willing to merge it.

devdazed:
i am also very interested in this.

pezra:
@devdazed, if you wanted to take this feature over that would be fine by me. Just make sure you reference this PR in the new one, if you do.

outoftime:
Hi from retirement! I think it would be really cool to model static columns as a flavor of parent-child relationships. Basically the static columns comprise the “parent” model, and queries against the table act like a precooked `JOIN` query. Just a thought!

outoftime:
(This could generalize to situations where a composite-keyed table has no static columns, but also does not have a corresponding “parent” table—in that case the parent model would just have a key and a `has_many` association)

pezra:
Hmm... @outoftime, are you suggesting something like:
```
class Child
  # ...
  belongs_to :parent, denormalize: [:column_x_from_parent, :column_y_from_parent]
```
Which would result in `parent_column_x` and `parent_column_y` static columns on `Child`. Querying `Child` would return model instances that always had a `parent` instance that was partially loaded with the parent id, column_x and column_y values from the child table?

That is in a interesting idea, though it sounds like quite a bit of work.

outoftime:
@pezra what I had in mind was a little different:

```ruby
class Parent
  include Cequel::Record::Static

  key :subdomain, :string
  column :column_x, :string
  column :column_y, :string
end

class Child
  include Cequel::Record

  belongs_to :parent
  key :id, :uuid
  column :column_z, :string
end
```

These two classes would together model a single underlying table, where the partition key is `subdomain`; the clustering column is `id`; `column_x` and `column_y` are static columns; and `column_z` is a normal column.

pezra:
That is better than what i imagined you where suggesting. :smile: I like it. I'd be ok with the more simplistic approach of this PR too, though.

pezra:
@ryanwilliams, @devdazed, et al, i am closing this PR due to inactivity. Please reopen if you can devote some time to completing this work.

PR #179, opened 2014-10-16
Touch up the connection retry code.

clarkbremer:
- Move retry description comments from #execute to
#execute_with_consistency.
- More paranoid check for max_retries.
- Re-order retry steps so that if max_retries == 0, we make no attempt
to recover.

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -161,11 +161,7 @@ module Cequel
       end
 
       #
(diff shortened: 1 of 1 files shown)


pezra:
Rather than this approach, i would prefer a way to provide a lambda that will be called on failure to handle recovery and decide whether or not to retry. The default lambda could provide the current functionality. This would provide the ability to do the rare/sophisticated error handling you mentioned without harming the more common use cases.

Issue #194, opened 2014-11-20
Using to_json with only filtering on cequel model object


zshenker:
Hi,

I have been trying to use to_json with a set of only options to filter the fields returned in the JSON, but am running in to some issues with cequel model objects.

i.e. I am running code like this: title.to_json(only: [:name, :account_id]

From what I have been able to see, the issue come in, in the serializeable_hash model, where the following code is run:
attribute_names = attributes.keys
if only = options[:only]
  attribute_names &= Array(only).map(&:to_s)

The values of the hashes here are as follows:
attribute_names [:updated_at, :name, :account_id, :id]
only [:name, :account_id]

It looks like this method is expecting the attribute.keys to return string values, rather than the symbols that are there at the moment.

I am not sure what the impact to anything else would be to change the attribute.keys to return strings rather than symbols? any ideas here?

Zac

PR #195, opened 2014-11-20
Return attribute key names as strings rather than symbols

zshenker:
See #194


outoftime:
I think [indifferent access](https://github.com/cequel/cequel/issues/210) is really the move here

outoftime:
Closing in favor of #231

Issue #46, opened 2013-10-01
Time zone support for timestamp columns


outoftime:
- [ ] In the Cequel::Metal layer, all timestamp column values in result rows should be converted to UTC
- [ ] In Cequel::Record, timestamp attributes should use the configured ActiveSupport time zone (i.e. `value.in_time_zone`)

There's no need to worry about time zones on write because Cassandra stores timestamps as milliseconds since epoch, so there's no attached time zone information at all.

PR #198, opened 2014-11-21
Calls in_time_zone on Time objects in Metal::Row

zshenker:
See #46

### DIFF
--- a/lib/cequel/metal/row.rb
+++ b/lib/cequel/metal/row.rb
@@ -24,7 +24,7 @@ module Cequel
                 if $1 == 'ttl' then row.set_ttl($2, value)
                 else row.set_writetime($2, value)
                 end
-              elsif value.respond_to?(:in_time_zone) row[name] = value.in_time_zone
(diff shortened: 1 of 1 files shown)


lib/cequel/metal/row.rb
```diff
@@ -24,6 +24,7 @@ def self.from_result_row(result_row)
                 if $1 == 'ttl' then row.set_ttl($2, value)
                 else row.set_writetime($2, value)
                 end
+              elsif value.respond_to?(:in_time_zone) && value.is_a?(Time) then row[name] = value.in_time_zone
```
outoftime: Do we need the `is_a?` check here? Seems like if it responds to `in_time_zone`, that should be enough for our purposes

outoftime: @zshenker :point_up:

pezra:
@zshenker, I am closing this due to lack of activity. Please reopen if you find the time to address @outoftime's comments.

PR #205, opened 2014-12-17
safe nil

kusakari:


outoftime:
@kusakari is there a specific situation in which a nil exception is being thrown because we're not checking for `nil` here? If there is, it's probably a bug higher up in the call chain—these methods should not be called unless the record set's current state allows it (and callers typically assume a non-nil return value).

kusakari:
@outoftime  Thank you very much for your reply.
I want you to see the below pattern.
```
class UserGift
  include Cequel::Record

  key :user_id, :uuid
  column :gift_id, :uuid, :index => true
  column :type_id, :int, default: 1
  timestamps
end
```

```
user_gifts = UserGift.where(user_id: user.id, gift_id: gift.id).to_a
NoMethodError: undefined method `name' for nil:NilClass
from /Users/xxxxx/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/bundler/gems/cequel-101b07cb8cdc/lib/cequel/record/record_set.rb:804:in `next_unscoped_key_name'
```

outoftime:
@kusakari I see—I do think we're doing something wrong higher in the call chain here; I will take a look.

kusakari:
Thanks!

PR #215, opened 2015-01-22
skip dir concerns under app/models for rake cequel:migrate

islue:
To avoid error as below:
LoadError: Unable to autoload constant Concerns::Dummy, expected .../test/app/models/concerns/dummy.rb to define it

cat dummy.rb
module Dummy
  extend ActiveSupport::Concern
  ...
end

Since rails autoloads app/models/concerns, files in the dir have no namespace Concerns.
http://guides.rubyonrails.org/constant_autoloading_and_reloading.html#autoload-paths

### DIFF
--- a/lib/cequel/record/tasks.rb
+++ b/lib/cequel/record/tasks.rb
@@ -42,7 +42,7 @@ namespace :cequel do
       new_constants.each do |class_name|
         begin
           clazz = class_name.constantize
-        rescue NameError, RuntimeError # rubocop:disable HandleExceptions
+        rescue LoadError, NameError, RuntimeError # rubocop:disable HandleExceptions
         else
           if clazz.is_a?(Class)
             if clazz.ancestors.include?(Cequel::Record) &&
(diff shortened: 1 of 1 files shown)


lib/cequel/record/tasks.rb
```diff
@@ -26,6 +26,7 @@
     project_root = defined?(Rails) ? Rails.root : Dir.pwd
     models_dir_path = "#{File.expand_path('app/models', project_root)}/"
     model_files = Dir.glob(File.join(models_dir_path, '**', '*.rb'))
+    model_files.reject! {|file| file =~ /^#{File.join(models_dir_path, 'concerns')}/}
```
outoftime: @islue is this necessary? I think the code below should already handle any bad situation that might pop up…

islue:
I digged it a little more. The problem occurs only under rails 4.2 with namespaced concerns modules.
I created an empty rails project with adding the file below only.

    $ cat app/models/concerns/myapp/foo.rb
    module Myapp
      module Foo
        extend ActiveSupport::Concern
      end
    end

under rails 4.1, rake cequel:migrate also throw en exception but ignored by rescue
#<RuntimeError: Circular dependency detected while autoloading constant Concerns::Myapp::Foo>

under rails 4.2, exception changed
#<LoadError: Unable to autoload constant Concerns::Myapp::Foo, expected /repos/github.com/islue/myapp/app/models/concerns/myapp/foo.rb to define it>

I updated the pull request.

Issue #193, opened 2014-11-18
Error running with Rails 4.2.0beta4


zshenker:
I receive the following error when trying to use cequel with Rails 4.2.0beta4:

/usr/local/lib/ruby/2.1.0/forwardable.rb:132:in `instance_delegate': wrong number of arguments (2 for 1) (ArgumentError)
	from /usr/local/bundle/gems/activemodel-4.2.0.beta4/lib/active_model/naming.rb:219:in `extended'
	from /usr/local/bundle/gems/activemodel-4.2.0.beta4/lib/active_model/validations.rb:42:in `extend'
	from /usr/local/bundle/gems/activemodel-4.2.0.beta4/lib/active_model/validations.rb:42:in `block in <module:Validations>'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/concern.rb:120:in `class_eval'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/concern.rb:120:in `append_features'
	from /usr/local/bundle/gems/cequel-1.4.5/lib/cequel/record/validations.rb:27:in `include'
	from /usr/local/bundle/gems/cequel-1.4.5/lib/cequel/record/validations.rb:27:in `block in <module:Validations>'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/concern.rb:120:in `class_eval'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/concern.rb:120:in `append_features'
	from /usr/local/bundle/gems/cequel-1.4.5/lib/cequel/record.rb:92:in `include'
	from /usr/local/bundle/gems/cequel-1.4.5/lib/cequel/record.rb:92:in `block in <module:Record>'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/concern.rb:120:in `class_eval'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/concern.rb:120:in `append_features'
	from /usr/src/app/app/models/asset.rb:3:in `include'
	from /usr/src/app/app/models/asset.rb:3:in `<class:Asset>'
	from /usr/src/app/app/models/asset.rb:1:in `<top (required)>'
	from /usr/local/bundle/gems/activesupport-4.2.0.beta4/lib/active_support/dependencies.rb:452:in `load'
[...]

PR #218, opened 2015-02-04
Fix Module#delegate when using Forwardable

outoftime:
Rails defines the `delegate` method directly on the `Module` class, meaning that `Forwardable#delegate` overrides it any time a class extends `Forwardable`.

So, define `Cequel::Util::Forwardable`, which mixes in vanilla `Forwardable` but then overrides the `delegate` method to use ActiveSupport's implementation defined directly on `Module`.

The overwriting of `#delegate` isn't a new problem in Rails 4.2, but Rails 4.2 introduces a new use of `#delegate` internally that breaks Cequel entirely. This fixes that breakage.


outoftime:
Fixes #193

PR #220, opened 2015-02-05
Fix cannot call original Forwardable#delegate issue.

Sixeight:
Thank you #218 .

Now, we can this without error (thanks!),

```ruby
class A
  include Cequel::Record

  attr_reader :b
  delegate :meth, to: :b
end
```

but we got a error with following code,

```ruby
class A
  include Cequel::Record

  attr_reader :b
  delegate :meth => :b
end
#=> `ArgumentError: Delegation needs a target. Supply an options hash with a :to key as the last argument (e.g. delegate :hello, to: :greeter).`
```

### DIFF
--- a/lib/cequel/util.rb
+++ b/lib/cequel/util.rb
@@ -53,14 +53,10 @@ module Cequel
     module Forwardable
       include ::Forwardable
 
-      def delegate_with_argument_check(*args, &block)
-        if args.size == 1
-          delegate_without_argument_check(args.first)
-        else
-          Module.instance_method(:delegate).bind(self).call(*args, &block)
-        end
+      def delegate(*args, &block)
+        return super if args.size == 1
+        Module.instance_method(:delegate).bind(self).call(*args, &block)
       end
-      alias_method_chain :delegate, :argument_check
     end
   end
 end
--- a/lib/cequel/util.rb
+++ b/lib/cequel/util.rb
@@ -53,9 +53,14 @@ module Cequel
     module Forwardable
       include ::Forwardable
 
-      def delegate(*args, &block)
-        Module.instance_method(:delegate).bind(self).call(*args, &block)
+      def delegate_with_argument_check(*args, &block)
+        if args.size == 1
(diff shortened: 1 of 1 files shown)


outoftime:
@Sixeight are you using the `Forwardable` implementation of `delegate`? Is it too much to just use the unambiguous `instance_delegate` instead?

I'm willing to be convinced that we need to support both…

lib/cequel/util.rb
```diff
@@ -53,9 +53,14 @@ def hattr_accessor(hash, *attributes)
     module Forwardable
       include ::Forwardable
 
-      def delegate(*args, &block)
-        Module.instance_method(:delegate).bind(self).call(*args, &block)
+      def delegate_with_argument_check(*args, &block)
+        if args.size == 1
+          delegate_without_argument_check(args.first)
```
outoftime: Can't we just call `super` here (and skip the `alias_method_chain`)?

lib/cequel/util.rb
```diff
@@ -53,9 +53,14 @@ def hattr_accessor(hash, *attributes)
     module Forwardable
       include ::Forwardable
 
-      def delegate(*args, &block)
-        Module.instance_method(:delegate).bind(self).call(*args, &block)
+      def delegate_with_argument_check(*args, &block)
+        if args.size == 1
```
outoftime: `args.one?` is more expressive

Sixeight:
Thank you for your comment.
I'm not using `Forwardable#delegate`, but I think that we should support both delegate for existing project.

lib/cequel/util.rb
Sixeight: We can just call `super`, thanks!

lib/cequel/util.rb
Sixeight: I did't know `Enumerable#one?` :astonished:

outoftime:
@Sixeight great, thanks!!

Sixeight:
@outoftime thank you!

PR #222, opened 2015-02-27
Handle camel cased methods or methods containing periods

kemper:
I was spiking using your gem for an app that has an existing table schema with columns that have camel cased names and names with periods in them. In CQL, these just need to be escaped with quotes. I've gone through and changed all the places that generate CQL that would need to be changed to handle this.

The associated specs I modified failed before I added the non-test code to fix them, let me know if other specs are needed.

This was tested with Cassandra 2.0.8.39.

### DIFF
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -605,7 +605,7 @@ module Cequel
       # @return [String] CQL `SELECT` statement encoding this data set's scope.
       #
       def cql
-        statement = Statement.new
+        Statement.new
           .append(select_cql)
           .append(" FROM #{table_name}")
           .append(*row_specifications_cql)
@@ -689,9 +689,9 @@ module Cequel
       end
 
       def select_cql
-        all_columns = select_columns +
-          ttl_columns.map { |column| "TTL(#{column})" } +
-          writetime_columns.map { |column| "WRITETIME(#{column})" }
+        all_columns = select_columns.map{|c| %("#{c}")} +
+          ttl_columns.map { |column| %(TTL("#{column}")) } +
+          writetime_columns.map { |column| %(WRITETIME("#{column}")) }
 
         if all_columns.any?
           "SELECT #{all_columns.join(',')}"
@@ -707,7 +707,7 @@ module Cequel
       def sort_order_cql
         if sort_order.any?
           order = sort_order
-            .map { |column, direction| "#{column} #{direction.to_s.upcase}" }
+            .map { |column, direction| %("#{column}" #{direction.to_s.upcase}) }
             .join(', ')
           " ORDER BY #{order}"
         end
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -689,7 +689,7 @@ module Cequel
       end
 
       def select_cql
-        all_columns = select_columns.map{|c| %("#{c}")} +
+        all_columns = select_columns.map { |c| %("#{c}") } +
           ttl_columns.map { |column| %(TTL("#{column}")) } +
           writetime_columns.map { |column| %(WRITETIME("#{column}")) }
 
--- a/lib/cequel/metal/deleter.rb
+++ b/lib/cequel/metal/deleter.rb
@@ -44,7 +44,7 @@ module Cequel
       #
       def list_remove_at(column, *positions)
         statements
-          .concat(positions.map { |position| "#{column}[#{position}]" })
+          .concat(positions.map { |position| %("#{column}"[#{position}]) })
       end
 
       #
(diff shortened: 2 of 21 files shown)


kemper:
Turns out this requires more changes than I thought. This commit allows fields to have special characters (and camel case fields), but not keys. I have some more changes that make my projects specs pass. I'm going to add some specs to this gem for those and add these extra changes to the pull request.

kemper:
I combed the codebase as many ways as I could think of in order to find all cases of the code interpolating column names without escaping and converted them. I did this initially with the specs against my codebase, but then went back and backfilled as needed and discovered a lot of other cases along the way.

I tried to minimize code changes in the specs so there wouldn't be as big a risk of having merge conflicts with other work, but it was tough given how this was a somewhat widespread change that I wanted to verify.

The trickiest part I think was the Finder method generation code. I think I ended up with an elegant replacement. In general, I find it preferable to use define_method anyway, so I consider this a win even though the number of lines of code increased. For reference on the topic:

http://tenderlovemaking.com/2013/03/03/dynamic_method_definitions.html

I don't see any reason to be concerned about define_method forming a closure over it's arguments; it should just be a small set of symbols that are interned along with some arrays in memory.

I considered adding a line before defining the finder method:

return if self.respond_to?(method_name)

since it appeared to me that the method definition was happening twice. However, since the specs are redefining classes repeatedly, I decided to assume I was getting tricked by seeing duplicate finder definitions occurring during my testing. I might revisit that line.

Also, there was an intermittently failing spec in spec/examples/record/persistence_spec.rb:259 that I fixed. I suspect some data was bleeding around the tests; It always passed before when run individually and when the file is run, but fails often when the all the specs are run.

clarkbremer:
:+1: 
I also prefer define_method()
Nice Work.

kemper:
I also confirmed that there were extra method definitions happening from the finders. It occurs for every key defined on a model after the first. This occurs because def_key_finders above calls def_finder twice, in order to define the latest key finder and the key1_and_key2 finders.

This last commit fixes that issue.

kemper:
I reverted the commit that prevents the superfluous finder definitions. It caused a couple of the spec/examples/spec_support/preparation_spec.rb spec's to fail on jruby. It was non-obvious why, but I guess it's not important. Defining methods multiple times doesn't cause any harm; it just creates a bit of confusion when it's encountered.

Let me know if you want me to squash or cleanup this pull request in some way.

outoftime:
@kemper thanks, this looks like it required a lot of effort, and it's clearly a change for the better. I have two overall concerns:

1. I think we need a `quote_column_name` method somewhere—as it stands in this PR the knowledge of how to quote a column name is duplicated many times over. As a corollary, is the quoting logic complete here? In particular, is it possible for a column name to contain a `"` character? (I can't readily find an answer to that question in the documentation)
2. I use dynamic source code evaluation in favor of `define_method` intentionally, because `define_method` causes the method definition to carry the additional baggage of the lexical context in which it was defined. I'd be happy to debate that convention on the merits, but let's do so in a separate pull request.

kemper:
For Point 1, I considered it, but I thought that might cause the code to be less readable. It certainly means that later changes will general require a change in one location vs many files. I think I fizzled a bit on trying to decided where that method would live. Suggestions?

For Point 2, I didn't choose define_method for stylistic reasons and I don't think it can be done separate of this pull request (the specs would not pass); I did it because if a column has a period in it, it's no longer a valid method name as far as statically defined method names are concerned. This is true of other cases, such as fields beginning with a capitalized letter or starting with a number. The options as far as I could tell were:

1. Detect all cases of field names that do not result in valid method names and do not generate those methods with eval.
2. Rescue the syntax error from class eval and continue on as though that is a sometimes expected case. Functionally similar to option 1, much more evil and potentially obscuring other problems, but not requiring the implementation of a code that validates ruby methods names.
3. Use define_method and allow methods that are not typically valid in ruby, but via define_method can be defined on objects.

Option 3 seemed preferable to me. Though I can't call those methods the normal way but they are still available to call via the bracket method e.g. - object["some.field"]

outoftime:
Hey Kris,

I definitely agree about the readability situation but in this case I think it's worth it to not have repeated knowledge scattered throughout the code. In fact, it looks like the driver gives us [a method to do the right thing](https://github.com/datastax/ruby-driver/blob/10769a65ae9eb758335bfde57d9b152c34817a8b/lib/cassandra/util.rb#L135). I also think it would make sense for `Column#to_s` to return the quoted name, which should cut down on some of the repetitive/difficult to read effects.

As for the second point, I hadn't thought about that—however it seems of exceedingly limited use to define methods that can't be called using the usual Ruby syntax. I'd propose instead normalizing column attribute names to be valid Ruby identifiers, e.g. replace all non-word characters with an underscore. Of course this could in theory result in non-unique attribute names generated from unique column names, but I think this highly unlikely in practice, and if anyone complains we can just provide an option to override the attribute name associated with a column.

What do you think?

pezra:
@kemper, if you implement @outoftime's suggestions above and rebase i would merge this PR.

pezra:
I am closing this due to inactivity. @kemper, if you want to finish this up please reopen it.

kemper:
Yeah, sorry. I'm not using Cassandra these days and haven't had time to
dust it off.

PR #227, opened 2015-03-26
Allow for Cassandra SSL configuration options

maxpospischil:
Not sure how to go about testing this as Cassandra can either enable or disable SSL and is not able to support both simultaneously. Would require a completely separate machine with client_encryption_ebabled: true

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -30,6 +30,8 @@ module Cequel
       attr_writer :default_consistency
       # @return [Hash] credentials for connect to cassandra
       attr_reader :credentials
+      # @return [Hash] SSL Configuration options
+      attr_reader :ssl_config
 
       #
       # @!method write(statement, *bind_vars)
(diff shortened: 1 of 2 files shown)


maxpospischil:
Okay I have added a test to make sure that the ssl config stuff is all extracted correctly for passing to cluster. I believe that this is sufficient because ruby-driver tests that the correct thing is done with those when they are passed into cluster.

lib/cequel/metal/keyspace.rb
```diff
@@ -266,6 +287,7 @@ def client_without_keyspace
       def client_options
         {hosts: hosts, port: port}.tap do |options|
           options.merge!(credentials) if credentials
+          options[:server_cert] = server_cert if server_cert
```
outoftime: It looks like `server_cert` is the only one of these configuration keys that we actually use anywhere?

At a higher level, I don't think it makes sense to have all of these as top-level values in the configuration. After all, it only makes sense to have `:server_cert`, `:private_key`, etc. if `:ssl` is `true`. I would propose rolling all the SSL-related values up into a single object (a `Struct` would probably do fine), and then failing fast if `:ssl` is set to true but any of the SSL-related properties are missing.

Note that this doesn't necessarily mean changing the interface to `#configure`; I'm not strongly opposed to having these as top-level keys in the options passed to that method, just as top-level instance variables in the `Keyspace` itself.

maxpospischil: Not quite true, you could have many different combinations. You could have only ssl: true, only server_cert, or all of server_cert, client_cert, private_key, and passphrase.... I have changed the extraction to place it all in a single hash that gets merged into the config if all values aren't nil. http://datastax.github.io/ruby-driver/features/security/ssl_encryption/ details the config options one could pass to Cassandra.cluster

PR #231, opened 2015-04-21
Fixing integration with ActiveModel::Serialization

HugoLnx:
The integration with ActiveModel::Serialization module was failing because it expects Model#attributes keys to be strings and Cequel::Record::Properties.attributes method return a hash with symbols as keys.

To fix that reducing the incompatibility between versions I changed the #attributes method to return a hash with indifferent access (as was advised by @outoftime in #195).

Because of this bug, the ``#as_json`` with ``:only`` option was returning ``{}`` always. 

Related issues : #194 #195.

### DIFF
--- a/lib/cequel/record/properties.rb
+++ b/lib/cequel/record/properties.rb
@@ -256,11 +256,11 @@ module Cequel
       end
 
       #
-      # @return [Hash<Symbol,Object>] map of column names to values currently
+      # @return [Hash<String,Object>] map of column names to values currently
       #   set on this record
       #
       def attributes
-        attribute_names.each_with_object({}) do |name, attributes|
+        attribute_names.each_with_object(HashWithIndifferentAccess.new) do |name, attributes|
           attributes[name] = read_attribute(name)
         end
       end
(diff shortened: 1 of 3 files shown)


lib/cequel/record/properties.rb
```diff
@@ -262,7 +262,7 @@ def attribute_names
       def attributes
         attribute_names.each_with_object({}) do |name, attributes|
           attributes[name] = read_attribute(name)
-        end
+        end.with_indifferent_access
```
outoftime: I think it would be a bit more efficient to just initialize a `HashWithIndifferentAccess` on line 263, rather than building up a normal `Hash` as an intermediate structure and then converting it to a `HashWithIndifferentAccess` here.

outoftime:
@HugoLnx general approach looks good, left one comment for how we could do it a little more efficiently

HugoLnx:
My bad, I already made the refactoring. :+1:

outoftime:
@hugolnx LGTM! Will merge/release next time I'm at a computer : )

PR #252, opened 2015-09-18
Adapt to Cassandra version to avoid CASSANDRA-8733

lesliev:
This Cassandra bug causes list updates to reverse the order of the
array sent, which was being worked around in List#unshift by
reversing the array that was sent.

This change detects the version of Cassandra and only reverses
when it's a known buggy version. Also, the tests detect the
version and expect a list_prepend operation to produce reversed
lists in buggy versions.

See: https://issues.apache.org/jira/browse/CASSANDRA-8733
The bug was fixed in 2.1.3 and 2.0.13.

This code depends on the .cassandra-versions file to compile
a list of buggy versions, but doesn't add any other dependencies.

The following three tests now pass:
 Cequel::Record::List#unshift should persist unloaded unshift
 Cequel::Record::List#unshift should atomically unshift
 Cequel::Metal::DataSet#list_prepend should prepend multiple elements to list column

Tested on the following versions of Cassandra:
 2.0.4
 2.0.12
 2.0.13
 2.0.16
 2.1.2
 2.1.3
 2.1.8
 2.2.1

### DIFF
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -187,8 +187,8 @@ module Cequel
       # @example
       #   posts.list_prepend(:categories, ['CQL', 'ORMs'])
       #
-      # @note If multiple elements are passed, they will appear in the list in
-      #   reverse order.
+      # @note A bug (CASSANDRA-8733) exists in Cassandra versions 0.3.0-2.0.12 and 2.1.0-2.1.2 which
+      #   will make elements appear in REVERSE ORDER in the list.
       # @note If a enclosed in a Keyspace#batch block, this method will be
       #   executed as part of the batch.
       # @see #list_append
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -254,6 +254,32 @@ module Cequel
         end
       end
 
+      # @return [String] Cassandra version number
+      def cassandra_version
+        return @cassandra_version if @cassandra_version
+
+        statement = <<-CQL
+          SELECT release_version
+          FROM system.local
+        CQL
+
+        log('CQL', statement) do
+          @cassandra_version = client_without_keyspace.execute(statement).first['release_version']
+        end
+      end
+
+      # return true if Cassandra server version is known to include bug CASSANDRA-8733
+      def bug8733_version?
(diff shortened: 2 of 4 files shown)


outoftime:
@lesliev thanks for this! I don’t actually see any need to adapt to the bug—better to keep the code simple and just do the right thing for the latest versions

lesliev:
Thanks for looking at this!

I totally agree with the keeping the code simple but the problem is that List#unshift currently works around the bug by calling .reverse. This means that for older Cassandra versions it works as you would expect - arrays are not reversed when you unshift/prepend them. If people use these methods, upgrading Cassandra at or beyond 2.0.13/2.1.3 would suddenly start corrupting data.

Taking the .reverse out of Cequel with an older Cassandra would make the data corruption start when Cequel was upgraded, up until Cassandra was upgraded too. I think that compounds the problem.

Since there's already a work-around in place, would it not be better to make the work-around intelligent, to keep the behaviour deterministic and protect users of the gem from this bug in Cassandra? The extra code to optionally reverse could be removed down the line when fewer people are using old versions.

DataSet#list_prepend is more low-level and I'd think it fairly unlikely that users of the gem would use this method - besides which the bug is documented there to a certain extent and my PR makes it even clearer.

Imagine you had a production system using List#unshift and rapidly inserting data - and you discovered this problem a few weeks after an upgrade. Fixing those lists would probably be impossible at that point.

lesliev:
I was thinking: you could be very loud about not supporting earlier versions of Cassandra with the next release of Cequel - but what upgrade path does that leave? Users would have to upgrade the database AND the gem at exactly the same time, which can be quite hard to do in production.

outoftime:
@lesliev OK, I’m convinced by your argument—let’s support all versions for now, and maybe down the line we can remove support for the buggy versions once we can be confident pretty much everyone has migrated.

lesliev:
Awesome!

lukes:
What's the status of this PR? Can it be merged into master?

lukes:
@lesliev The master cequel branch has had its failing tests fixed, so if you rebase with it and re-push this branch then we might see your PR accepted .. :) cc @outoftime

lesliev:
@lukes done!

PR #254, opened 2015-09-30
Fix cql/3.0 broken link as new is cql/3.3

elgalu:
Still missing 2 links to fix, help needed.


pezra:

PR #267, opened 2016-01-08
syntatic sugar for reseting a keyspace

orenmazor:
This has been driving me crazy.


orenmazor:
nvm looks like this exists under a different namespace :+1: :+1:

Issue #142, opened 2014-08-18
Cql::Client.connect options


twoofy:
Cql::Client.connect supports options that Cequel seems to be ignoring.    The ones we specifically are using (some of which could/should probably be removed) are: connection_timeout, cql_version, and default_consistency.

I believe the issue (if there is one) is due to this code in lib/cequel/metal/keyspace.rb:

```
  def client_options
    {hosts: hosts, port: port}.tap do |options|
      options[:credentials] = credentials if credentials
    end
  end
```

Any plans to support more of the connect options in Cql::Client.connect?

cmbaron:
I took a stab at exposing ssl options based on https://gist.github.com/jasonmk/be5257c2b846c2b74c6c 
Any advice on how to build a test for https://github.com/cmbaron/cequel/commit/a342e1d6fde6eef8f2131dcf2dbfc5cd8e165e9c given cassandra does not seem to support ssl and non-ssl concurrently?

bobbrez:
It seems like `auth_provider` may be needed in cases of using [Instaclustr on Heroku](https://devcenter.heroku.com/articles/instaclustr#using-with-ruby). Testing this out on a project, using a config file like

```yaml
production:
  hosts: <%= ENV['IC_CONTACT_POINTS'].to_s.split(',') %>
  port: <%= ENV['IC_PORT'] %>
  keyspace: atlas_production
  username: <%= ENV['IC_USER'] %>
  password: <%= ENV['IC_PASSWORD'] %>
  max_retries: 3
  retry_delay: 0.5
  newrelic: true
  replication:
    class: NetworkTopologyStrategy
    replication_factor: 1
    datacenter1: US_EAST_1
```
Will throw `Cassandra::Errors::AuthenticationError: Server requested authentication, but client was not configured to authenticate` on any attempt to connect to the cluster, but using an adapter resolves the issue. I'll try to put together a PR for this.

raffe90:
@bobbrez I keep getting error: 

`Cassandra::Errors::AuthenticationError: Server requested authentication, but client was not configured to authenticate`

How did you fix this?

bobbrez:
@raffe90 I just submitted a PR to this project with my fix. Please take a look and improve it as necessary. I've been using this in production since Sept without any issue.

PR #274, opened 2016-02-04
Enable setting an auth_provider configuration

bobbrez:
As per issue #142 this allows setting an auth_provider

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -316,7 +316,12 @@ module Cequel
       end
 
       def extract_credentials(configuration)
-        configuration.slice(:username, :password).presence
+        credentials = configuration.slice(:username, :password).presence
+        
+        return credentials unless configuration[:auth_provider]
+
+        { auth_provider: CequelPassword.new(*credentials.values),
+          datacenter: configuration[:datacenter].presence }
       end
 
       def extract_max_retries(configuration)
@@ -336,6 +341,12 @@ module Cequel
         ssl_config[:passphrase] = configuration.fetch(:passphrase, nil)
         ssl_config.each { |key, value| ssl_config.delete(key) unless value }
         ssl_config
+      end
+    end
+
+    class CequelPassword < Cassandra::Auth::Providers::Password
+      def create_authenticator(_)
+        Authenticator.new(@username, @password)
       end
     end
   end


lib/cequel/metal/keyspace.rb
```diff
@@ -316,7 +316,12 @@ def extract_hosts_and_port(configuration)
       end
 
       def extract_credentials(configuration)
-        configuration.slice(:username, :password).presence
+        credentials = configuration.slice(:username, :password).presence
+        
+        return credentials unless configuration[:auth_provider]
+
+        { auth_provider: CequelPassword.new(*credentials.values),
+          datacenter: configuration[:datacenter].presence }
```
pezra: This seems to ignore the value of `configuration[:auth_provider]` which would surprise me, as a user.

Also, if we start supporting auth providers perhaps we should just go whole hog and always use an auth provider, even when `:username` and `:password` are specified in the configuration. Maybe something like

```
def extract_auth_provider
  auth_provider = configuration.fetch(:auth_provider, Cequel:Auth:Provider:Password)
  auth_provider.new(configuration)
end
```
```
class Cequel:Auth:Provider:Password
  def initialize(config)
    @user = config.fetch(:username)
    @pass = config.fetch(:password)
  end
  def create_authenticator(_)
    Cassandra::Auth::Providers::Password::Authenticator.new(@user, @pass)
  end
end

**Comment by pezra** [comment]
@bobbrez, if you rebase from master the test should pass.

Also, is there a reasonable way to test this pluggable auth behavior, do you think?

File: lib/cequel/metal/keyspace.rb (hunk 1 above)
Comment (bobbrez) [inline lib/cequel/metal/keyspace.rb]: That seems pretty reasonable. I went down this route because I didn't want to force the behavior, unsure if always including a provider would break an expectation. If its not a concern that I think its the right way to go.

**Comment by bobbrez** [comment]
My first instinct is to look at how something like [Faraday](https://github.com/lostisland/faraday/blob/master/test/middleware_stack_test.rb) does testing. It may be

**Comment by pezra** [comment]
@bobbrez, it is worth noting `configuration` comes in from a yaml file so the `extract_auth_provider` method needs to start from simple types and coerce them into the appropriate objects. Eg, rather than the code i wrote in my example i think it would need to read more like
```
configuration.fetch(:auth_provider, "Cequel:Auth:Provider:Password").constantize
```

pezra:
i am closing this due to inactivity. @bobbrez, if you want to finish this up please reopen.

PR #277, opened 2016-02-18
Timestamp cast to milliseconds

lukes:
Cassandra stores timestamps to a precision of milliseconds.

http://cassandra.apache.org/doc/cql3/CQL.html#usingtimestamps

Whereas Ruby allows Time objects to be more precise.

Before this change the timestamp value would often be different between
a model before it had been saved, and the same model once loaded from
Cassandra.

This would mean that the following would happen:

```ruby
published_at = Time.now

post = Post.find(1)
post.published_at = published_at
post.save

post = Post.find(1)
post.published_at = published_at
post.changes.present? # => true
```

Also, previously the check for if an attribute had changed was being
conducted on a non-casted value.

I've included two tests that previously would have failed.


pezra:
@lukes, if you rebase against master you should be able to get the specs passing. I'd be delighted to have this improvement once the tests are passing.

lukes:
@pezra Thank you for maintaining this gem! Tests are passing if you would like to review this PR again.

spec/examples/record/lazy_record_collection_spec.rb
```diff
@@ -8,7 +8,7 @@
       column :value, :text
     end
 
-    let(:now) { Time.now }
+    let(:now) { Time.now.round(3) }
     let(:event) { Event[now] }
 
     it { expect(event.attributes).to include(timestamp: now) }
```
pezra: I dislike the explicit rounding. What about remove the round and doing something like:
```
... expect(event.timestamp).to be_with(1.millisecond).of(now) ...
```

lukes:
Good call @pezra, I've updated the specs to be in line with your suggestion. I've opted for making a spec helper method `one_millisecond`, instead of creating any ActiveModel dependencies.

spec/examples/record/lazy_record_collection_spec.rb
```diff
@@ -11,6 +11,6 @@
     let(:now) { Time.now }
     let(:event) { Event[now] }
 
-    it { expect(event.attributes).to include(timestamp: now) }
+    it { expect(event.timestamp).to be_within(one_millisecond).of(now) }
```
pezra: :+1: for `one_millisecond`

pezra:
Merged. Thanks for the improvement!

lukes:
You're welcome, thanks for looking after this gem!

PR #278, opened 2016-02-18
Timestamp cast to milliseconds

lukes:
Cassandra stores timestamps to a precision of milliseconds.

http://cassandra.apache.org/doc/cql3/CQL.html#usingtimestamps

Whereas Ruby allows Time objects to be more precise.

Before this change the timestamp value would often be different between
a model before it had been saved, and the same model once loaded from
Cassandra.

This would mean that the following would happen:

```ruby
published_at = Time.now

post = Post.find(1)
post.published_at = published_at
post.save

post = Post.find(1)
post.published_at = published_at
post.changes.blank? # => false
```

Also, previously the check for if an attribute had changed was being
conducted on a non-casted value.

I've included two tests that previously would have failed.

### DIFF
--- a/lib/cequel/record/dirty.rb
+++ b/lib/cequel/record/dirty.rb
@@ -56,6 +56,10 @@ module Cequel
       private
 
       def write_attribute(name, value)
+        column = self.class.reflect_on_column(name)
+        fail UnknownAttributeError, "unknown attribute: #{name}" unless column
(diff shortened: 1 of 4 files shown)


lukes:
Sorry, was supposed to PR this into a different repo. Sorry for making a dupe.

Issue #272, opened 2016-01-20
partition_columns.sort { |c| c['component_index'] } is causing migration to fail on second migrations due to mismatch between partition keys  and types 


rwwaskk:
Hi Cequel Team, 

First of all thanks for the great gem 

Recently after our team upgraded from cequel 1.4.5 to cequel 1.7.0. This caused our subsequent migrations after the initial one to fail due to errors like "Cassandra::Errors::InvalidError: Cannot rename column id to date in keyspace; another column of that name already exist". This is NOT because we changed the column names. In fact, the cequel models has not been changed at all. 

After some debugging, I realized the problem comes from here in the table_reader:
```ruby
 def read_partition_keys
-        partition_columns.each do |column|
-          name, type = column.values_at("column_name", "validator")
+        validators = table_data['key_validator']
+        types = parse_composite_types(validators) || [validators]
+        columns = partition_columns.sort { |c| c['component_index'] }
+          .map { |c| c['column_name'] }
+
+        columns.zip(types) do |name, type|
           table.add_partition_key(name.to_sym, Type.lookup_internal(type))
         end
       end
```
The sort is not returning the correct ordering based on component index. 
We changed  partition_columns.sort to  partition_columns.sort_by and the problem is gone. 

Maybe I am not really understanding what the code does exactly. But I checked our database with this query,
```
SELECT column_name, index_name, index_options, index_type, component_index 
FROM system.schema_columns 
WHERE keyspace_name='samplekp'AND columnfamily_name='sampletable';
```
and it seems partition_columns has the correct component index return for each partition key. But the sort messes up the order and matches it with the wrong type.

Appreciate  your time for checking on this.

pezra:
fixed by #279

PR #279, opened 2016-02-20
fix sorting partition columns

tarhashi:
Hi.
I encountered the same problem... https://github.com/cequel/cequel/issues/272


pezra:
@tarhashi, if you rebase against master the specs should pass. I'd be delighted to have this functionality once the tests are green.

pezra:
Merged. Thanks for the fix.

PR #283, opened 2016-02-26
Page size and paging state

bettse:
The Cassandra driver supports pagination through a `page_size` and `paging_state`, and this exposes them on the record_set class, much like consistency is, so they can be used.

I haven't written tests for this or run the vagrant tests yet, I wanted to get some initial feedback on if I was going about this in right way beforehand.

### DIFF
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -580,7 +580,7 @@ module Cequel
       end
 
       def next_paging_state
-        execute_cql(*cql).paging_state
+        results.paging_state
       end
 
       # rubocop:enable LineLength
@@ -600,8 +600,7 @@ module Cequel
       #
       def each
         return enum_for(:each) unless block_given?
-        result = execute_cql(*cql)
-        result.each { |row| yield Row.from_result_row(row) }
+        results.each { |row| yield Row.from_result_row(row) }
       end
 
       #
@@ -676,6 +675,10 @@ module Cequel
       attr_writer :row_limit, :query_consistency, :query_page_size, :query_paging_state
 
       private
+
+      def results
+        @results ||= execute_cql(*cql)
+      end
 
       def execute_cql(cql, *bind_vars)
         keyspace.execute_with_options(cql, bind_vars, {
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -49,6 +49,7 @@ module Cequel
       # @since 1.1.0
       attr_reader :query_consistency
       attr_reader :query_page_size
+      attr_reader :query_paging_state
 
       def_delegator :keyspace, :write_with_consistency
 
@@ -572,6 +573,11 @@ module Cequel
         end
       end
 
+      def paging_state(paging_state)
+        clone.tap do |data_set|
+          data_set.query_paging_state = paging_state
+        end
+      end
 
       # rubocop:enable LineLength
 
@@ -663,12 +669,16 @@ module Cequel
 
       protected
 
-      attr_writer :row_limit, :query_consistency, :query_page_size
+      attr_writer :row_limit, :query_consistency, :query_page_size, :query_paging_state
 
       private
 
       def execute_cql(cql, *bind_vars)
-        keyspace.execute_with_options(cql, bind_vars, {consistency: query_consistency, page_size: query_page_size})
+        keyspace.execute_with_options(cql, bind_vars, {
+          consistency: query_consistency,
+          page_size: query_page_size,
+          paging_state: query_paging_state
+        })
       end
 
       def inserter(&block)
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -579,6 +579,10 @@ module Cequel
         end
       end
 
+      def next_paging_state
+        execute_cql(*cql).paging_state
+      end
+
       # rubocop:enable LineLength
 
       #
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -48,6 +48,7 @@ module Cequel
       #   use
       # @since 1.1.0
       attr_reader :query_consistency
+      attr_reader :query_page_size
 
       def_delegator :keyspace, :write_with_consistency
 
@@ -565,6 +566,13 @@ module Cequel
         end
       end
 
+      def page_size(page_size)
+        clone.tap do |data_set|
+          data_set.query_page_size = page_size
+        end
+      end
+
+
       # rubocop:enable LineLength
 
       #
@@ -655,12 +663,12 @@ module Cequel
 
       protected
 
-      attr_writer :row_limit, :query_consistency
+      attr_writer :row_limit, :query_consistency, :query_page_size
 
       private
 
       def execute_cql(cql, *bind_vars)
-        keyspace.execute_with_consistency(cql, bind_vars, query_consistency)
+        keyspace.execute_with_options(cql, bind_vars, {consistency: query_consistency, page_size: query_page_size})
       end
 
       def inserter(&block)
--- a/spec/examples/record/record_set_spec.rb
+++ b/spec/examples/record/record_set_spec.rb
@@ -811,6 +811,41 @@ describe Cequel::Record::RecordSet do
     end
   end
 
+  describe '#page_size' do
+    it 'should return the number of records specified by page_size' do
+      expect(Post.page_size(2).to_a.length).to be(2)
+    end
+  end
+
+  describe '#next_paging_state' do
+    it 'should return the paging state of the result' do
+      expect(Post.page_size(1).next_paging_state).not_to eq(nil)
+    end
+  end
+
+  describe '#paging_state' do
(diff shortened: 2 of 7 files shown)


lib/cequel/metal/data_set.rb
```diff
@@ -655,12 +673,16 @@ def row_specifications_cql
...
 
       private
 
       def execute_cql(cql, *bind_vars)
-        keyspace.execute_with_consistency(cql, bind_vars, query_consistency)
+        keyspace.execute_with_options(cql, bind_vars, {
+          consistency: query_consistency, 
+          page_size: query_page_size, 
+          paging_state: query_paging_state
+        })
```
pezra: Should query paging state be updated before returning the results?

pezra:
@bettse, I like the idea of using the native paging. If you rebase against master you should be able to get the specs passing. I'd be delighted to have this improvement once the tests are passing (and there are sometests for this new behavior).

lib/cequel/metal/data_set.rb
bettse: It could, and I think it would clean up (or make unnecessary) the `next_paging_state` if it did, but there is also the question of how the library user would know if the paging_state they're getting is the one they supplied vs the new one.  To some extent, I think that's why these are all prefixed with 'query', they're parameters of the query, not the result.

This is my first foray into the code, so there is a lot of speculation.  I can certainly refactor to have it update if that is the preferred style.

bettse:
Awesome.  I was kinda concerned the project had gone dormant.  I'll do that, and add a comment when I feel its passing the tests (has some new tests), and is ready to be looked at.  I think there are some places in the batch classes that I also noticed after making the PR and need to be updated.

lib/cequel/metal/data_set.rb
pezra: Hmm... I think i see your point. I would be helpful, for me, to see some examples of expected usage pattern. That would clarify the user experience this would create. A test or two might do the trick.

pezra:
@bettse, there was a bit of a lull in this project but i think we are back in business now. I look forward to seeing the final PR.

bettse:
I seem to be a bit stuck.  Travis shows one of the new tests I wrote failing on all versions, but locally against ruby 2.2.2 it was passing, and I've tried dropping the cequel_schema_test keyspace to see if it was a caching issue, and installed ruby 2.1.5 to try to more closely reproduce one of the travis runs.  I just `bundle exec rake test:all` and got success for all non-missing combinations:

```
Results:
MISSING gemfile: gemfiles/rails_4.0.gemfile rvm: 2.2.0        
MISSING gemfile: gemfiles/rails_4.0.gemfile rvm: jruby-9.0.0.0
MISSING gemfile: gemfiles/rails_4.1.gemfile rvm: 2.2.0        
MISSING gemfile: gemfiles/rails_4.0.gemfile rvm: jruby-1.7.22 
MISSING gemfile: gemfiles/rails_4.1.gemfile rvm: 2.3.0        
MISSING gemfile: gemfiles/rails_4.0.gemfile rvm: 2.3.0        
MISSING gemfile: gemfiles/rails_4.1.gemfile rvm: jruby-1.7.22 
MISSING gemfile: gemfiles/rails_4.1.gemfile rvm: jruby-9.0.0.0
MISSING gemfile: gemfiles/rails_4.2.gemfile rvm: 2.2.0        
MISSING gemfile: gemfiles/rails_4.2.gemfile rvm: 2.3.0        
MISSING gemfile: gemfiles/rails_4.2.gemfile rvm: jruby-1.7.22 
MISSING gemfile: gemfiles/rails_4.2.gemfile rvm: jruby-9.0.0.0
SUCCESS gemfile: gemfiles/rails_4.1.gemfile rvm: 2.1.5        
SUCCESS gemfile: gemfiles/rails_4.0.gemfile rvm: 2.1.5        
SUCCESS gemfile: gemfiles/rails_4.2.gemfile rvm: 2.1.5
```

I'm open to any ideas about how I might reproduce the failure locally, or if there are nuances to travis that could be playing a part (I have no experience with it).

bettse:
reproduced!  I had been testing against cassandra 2.0.4 since that was the example in the contributing file, but after bringing up a vagrant image for 2.1.3 (arbitrarily chosen), the test now fails and I can get to fixing it :)

bettse:
Green and ready for review!

lib/cequel/metal/data_set.rb
```diff
@@ -565,6 +567,22 @@ def consistency(consistency)
...
+
+      def paging_state(paging_state)
+        clone.tap do |data_set|
+          data_set.query_paging_state = paging_state
+        end
+      end
+
+      def next_paging_state
+        execute_cql(*cql).paging_state
+      end
```
pezra: Does this mean that the query is run twice? Ie, once to figure the paging state and once to iterate over the results? We should extract `execute_cql(*cql)` in to a `#results` method that memoizes that operation so we only do it once per data set instance.

spec/examples/record/record_set_spec.rb
```diff
@@ -811,6 +811,46 @@ def self.latest(count)
...
+
+  describe '#next_paging_state' do
+    it 'should return the paging state of the result' do
+      expect(Post.page_size(1).next_paging_state).not_to eq(nil)
+    end
+  end
+
+  describe '#paging_state' do
+    let(:page_one) { Post.page_size(1) }
+    let(:page_two) { Post.page_size(1).paging_state(page_one.next_paging_state) }
```
pezra: out of curiosity, what does a paging state look like?

spec/examples/record/record_set_spec.rb
```diff
@@ -811,6 +811,46 @@ def self.latest(count)
...
+      expect_query_with_options(/SELECT/, :paging_state => page_one.next_paging_state) do
+        expect(page_two.to_a.first).to eq(posts.to_a.second)
+      end
+    end
+
+    it 'should be nil when omitted' do
+      expect(Post.page_size(1).paging_state(nil).to_a).to eq(page_one.to_a)
+    end
+  end
+
```
pezra: Can we have a more behavior oriented test that retrieves all the records over however many pages there are and checks that we got all the expected values? I see the value of the preceding tests from a development perspective but from a maintenance perspective they check quite a bit of implementation detail (eg, the names of the paging control options to the cequel driver) that isn't really important to clients of this class. If those implementation details change in the future. these tests should continue to pass if the over all behavior is preserved.

spec/support/helpers.rb
```diff
@@ -139,6 +139,14 @@ def expect_query_with_consistency(matcher, consistency)
...
           with(matcher, hash_including(:consistency => consistency))
       end
+
+      def expect_query_with_options(matcher, options)
+        allow(cequel.client).to receive(:execute).and_call_original
+        yield
+        expect(cequel.client).to have_received(:execute).
+          with(matcher, hash_including(options))
+      end
+
```
pezra: :+1: for extracting test boiler plate into a helper

spec/examples/record/record_set_spec.rb
bettse: It depends on the structure of the table, but the one returned by page_one looks like this:

`"\x00\tcassandra\x00\x12\x00\acequel0\x00\x00\x05title\x00\x7F\xFF\xFF\xFE"`

Since it contains data about the table, the documentation for the ruby driver cautions that it should be encrypted:

http://www.rubydoc.info/gems/cassandra-driver/2.0.0/Cassandra/Execution/Options:paging_state

lib/cequel/metal/data_set.rb
bettse: You are correct, the query would be called a second time.  I can take a whack at the memoization.

spec/support/helpers.rb
bettse: Now that I look at it, I could probably clean up the method that checks for consistency to reuse this one as well.

spec/examples/record/record_set_spec.rb
bettse: Sure.  I'm not heavily wed to the tests as they are.  Its probably obvious I don't have much experience with library development, I think I was trying to do too many things at once.  

I wanted some independence to the tests, but really the page_size/paging_state/next_paging_state are very codependent.  I was also trying to fit in with how I saw the consistency tests being done since the page_size and paging_state follow similar paths.  Really the primary thing is to demonstrate that the use of the page_size/paging_state are effective at paginating.  

It sounds like that's more of what you're looking for: use the page_size/paging_state to page through the records(say, in a loop), validate that the pages are of the correct size, and that the collected pages are the correct set.  And it wouldn't do any of the 'expect_query_with_options' checks.  Sound about right?

pezra: That sounds exactly right.

pezra:
This is looking real close. A bit more polish and we'll get it merged.

bettse:
There were some intermittent test failures on travis, but now its all green. Let me know what you think :microscope:

pezra:
Merged. Thanks for putting that together.

bettse:
Thank you for your guidance!

orenmazor:
Finding this PR literally made my day.

PR #284, opened 2016-02-29
Attributes marked as dirty when assigned on a new object.

lukes:
This change brings Cequel models into line with regular ActiveRecord
model behaviour.

```ruby
p = Post.new(title: 'My Post')
p.changes # => { 'title' => [nil, 'My Post'] }
```

Previously, this would only happen on objects that had been persisted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
+sudo: false
 language: ruby
 rvm:
   - "2.2.0"
   - "2.1.5"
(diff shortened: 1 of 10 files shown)


pezra:
@lukes, if you rebase against master you should be able to get the specs passing. I'd be delighted to have this improvement once the tests are passing.

pezra:

PR #285, opened 2016-03-02
Don't assume next_unscoped_key_column has a value

Geesu:
When writing specs we were receiving `undefined method 'name' for nil:NilClass` which was hiding the true problem.

Once applying the fix in this PR we were able to see the proper exception/problem, specifically (in our case):
`Can't scope key column <our column> without also scoping `

### DIFF
--- a/lib/cequel/record/record_set.rb
+++ b/lib/cequel/record/record_set.rb
@@ -819,7 +819,7 @@ module Cequel
       end
 
       def next_unscoped_key_name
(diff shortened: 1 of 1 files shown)


pezra:
@Geesu, if you rebase against master the specs should pass. I'd be delighted to have this improvement  once the tests are passing.

pezra:
@Geesu, i am closing this due to inactivity. Feel free to reopen if you get a chance to work on this.

PR #287, opened 2016-03-10
Added :foreign_key option to belongs_to association. 

jhannus:
This PR adds a `:foreign_key` option to the `belongs_to` association.

This works well for: 
* Using cequel on schemas that do not follow a rails relational convention of naming or, that use a consistent naming throughout.
* Tables that have composite partition keys where parts are outside of your own domain i.e. `account_id` where `account_id` is used in many tables.  

Given the existing tables:
```sql
Create table parent (
  account_id uuid,
  parent_id uuid,
  name text,
  PRIMARY KEY (account_id, parent_id)
);

Create table child (
  parent_id uuid,
  child_id uuid,
  name text,
  PRIMARY KEY ((parent_id), child_id)
);
```

Example:
```ruby
class Parent
  include Cequel::Record
  key :account_id, :uuid, partition: true
  key :parent_id, :uuid, partition: true
  column :name, :text
end

class Child
  include Cequel::Record
  belongs_to :parent, foreign_key: [:account_id, :parent_id]
  key :child_id, :uuid, auto: true
  column :name, :text
end
```

### DIFF
--- a/lib/cequel/record/associations.rb
+++ b/lib/cequel/record/associations.rb
@@ -110,13 +110,13 @@ module Cequel
           end
 
           key_options = options.extract!(:partition)
-          foreign_key_parts = [options.fetch(:foreign_key, [])].flatten
 
           self.parent_association =
             BelongsToAssociation.new(self, name.to_sym, options)
 
           parent_association.association_key_columns.each_with_index do |column, i|
-            foreign_key = foreign_key_parts[i].nil? ? "#{name}_#{column.name}" : foreign_key_parts[i]
+            foreign_key_parts = self.parent_association.foreign_keys
+            foreign_key = foreign_key_parts.any? ? foreign_key_parts[i] : "#{name}_#{column.name}"
             key foreign_key.to_sym, column.type, key_options
           end
           def_parent_association_accessors
--- a/lib/cequel/record/associations.rb
+++ b/lib/cequel/record/associations.rb
@@ -110,12 +110,13 @@ module Cequel
           end
 
           key_options = options.extract!(:partition)
+          foreign_key_parts = [options.fetch(:foreign_key, [])].flatten
 
           self.parent_association =
             BelongsToAssociation.new(self, name.to_sym, options)
 
-          parent_association.association_key_columns.each do |column|
-            foreign_key = options.fetch(:foreign_key, "#{name}_#{column.name}")
+          parent_association.association_key_columns.each_with_index do |column, i|
+            foreign_key = foreign_key_parts[i].nil? ? "#{name}_#{column.name}" : foreign_key_parts[i]
             key foreign_key.to_sym, column.type, key_options
           end
           def_parent_association_accessors
--- a/lib/cequel/record/associations.rb
+++ b/lib/cequel/record/associations.rb
@@ -108,14 +108,15 @@ module Cequel
                  "belongs_to association must be declared before declaring " \
                  "key(s)"
           end
-          
+
           key_options = options.extract!(:partition)
 
           self.parent_association =
             BelongsToAssociation.new(self, name.to_sym, options)
 
           parent_association.association_key_columns.each do |column|
-            key :"#{name}_#{column.name}", column.type, key_options
+            foreign_key = options.fetch(:foreign_key, "#{name}_#{column.name}")
+            key foreign_key.to_sym, column.type, key_options
           end
           def_parent_association_accessors
         end
--- a/lib/cequel/record/belongs_to_association.rb
+++ b/lib/cequel/record/belongs_to_association.rb
@@ -31,7 +31,7 @@ module Cequel
       # @api private
       #
       def initialize(owner_class, name, options = {})
-        options.assert_valid_keys(:class_name)
+        options.assert_valid_keys(:class_name, :foreign_key)
 
         @owner_class, @name = owner_class, name.to_sym
         @association_class_name =
--- a/lib/cequel/record/belongs_to_association.rb
+++ b/lib/cequel/record/belongs_to_association.rb
@@ -17,6 +17,8 @@ module Cequel
       attr_reader :name
       # @return [String] name of parent class
       attr_reader :association_class_name
+      # @return [Array] array of foreign key symbols
+      attr_reader :foreign_keys
 
       # @!attribute [r] association_key_columns
       #   @return [Array<Schema::Column>] key columns on the parent class
@@ -33,6 +35,7 @@ module Cequel
       def initialize(owner_class, name, options = {})
         options.assert_valid_keys(:class_name, :foreign_key)
 
+        @foreign_keys = Array(options.fetch(:foreign_key, [])).map { |x| x.to_sym }
         @owner_class, @name = owner_class, name.to_sym
         @association_class_name =
           options.fetch(:class_name, @name.to_s.classify)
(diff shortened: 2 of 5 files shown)


jhannus:
Looks like Travis is having some trouble building. I doesn't appear to be a result of my additions. Can someone help me understand the failures? Thanks!

jhannus:
@outoftime Can someone please help me with this PR. My apologies If this is not the right channel/person to ask. Thanks.

pezra:
@jhannus, if you rebase against master the specs should pass. I'd be delighted to have this functionality once the tests are passing.

jhannus:
Thanks @pezra All the specs are passing now.

pezra:
@jhannus, after thinking about it a bit (i'm a little new to the admin side of this project) i wonder if it would be be better to have an option that would just allow the user to specify the parent association key column name wholesale. Eg,

```
class Child
  include Cequel::Record
  belongs_to :parent, foreign_key: :parent_id
  key :child_id, :uuid, auto: true
  column :name, :text
end
```

This not only solves the specific scenario you encountered but a variety of others as well.

I am not tied to the `foreign_key` name but this seems analogous to what that option does in active record.

jhannus:
Thats an awesome suggestion. I'll make the changes and update the PR. Thanks @pezra !

jhannus:
@pezra the `foreign_key:` option now allows for arrays in the case where you have a composite partition key in the parent class.

lib/cequel/record/associations.rb
```diff
@@ -108,14 +108,16 @@ def belongs_to(name, options = {})
                  "belongs_to association must be declared before declaring " \
                  "key(s)"
           end
-          
+
           key_options = options.extract!(:partition)
+          foreign_key_parts = [options.fetch(:foreign_key, [])].flatten
```
pezra: Not to be overly picky but how about `... = Array(options[:foreign_key])`?

pezra: Or even `... = Array(options.fetch(:foreign_key, figure_default_foreign_key(parent_association)))`

spec/examples/record/associations_spec.rb
```diff
@@ -186,6 +186,51 @@
...
+
+        it "should add parent's keys as first keys" do
+          expect(Bar.key_column_names).to eq([:account_id, :parent_id, :child_id])
+        end
+
+        it "should add parent's keys as partition keys" do
+          expect(Bar.partition_key_column_names).to eq([:account_id, :parent_id])
+        end
+      end
+    end
```
pezra: :+1: for the test coverage!

lib/cequel/record/associations.rb
```diff
@@ -108,14 +108,16 @@ def belongs_to(name, options = {})
...
 
           self.parent_association =
             BelongsToAssociation.new(self, name.to_sym, options)
 
-          parent_association.association_key_columns.each do |column|
-            key :"#{name}_#{column.name}", column.type, key_options
+          parent_association.association_key_columns.each_with_index do |column, i|
+            foreign_key = foreign_key_parts[i].nil? ? "#{name}_#{column.name}" : foreign_key_parts[i]
+            key foreign_key.to_sym, column.type, key_options
```
pezra: Would it be better to make the `:foreign_key` option all or nothing? It seems likely if one column uses a different convention then all the others will too.

lib/cequel/record/associations.rb
jhannus: Never too picky! I'll take your suggestion and come up with something.

lib/cequel/record/associations.rb
jhannus: Can you please elaborate on 'all or nothing'? I guess I only added the array because partition keys can be composite.

jhannus: I updated the example in markdown above to include an attribute called account_id (on Parent) which is outside of our domain. This illustrates the use of the array foreign_key.

lib/cequel/record/associations.rb
jhannus: Oh and for the record i was never that happy with the `flatten` :) Probably wouldn't have passed review where I work either! LOL

lib/cequel/record/associations.rb
pezra: It is closely related to the comment above. As i read this code it seems that someone could write `... foreign_key: [nil, :parent_id]` and it would come out as `[:foo_account_id, :parent_id]`. That seems a little weird.

It would be better, in my view, to require the `:foreign_key`  to be either entirely absent or fully formed and complete.

jhannus:
@pezra Sorry for my late reply I've been busy here at work but will get to this shortly. Thanks!

jhannus:
@pezra I took your suggestion and also moved the foreign key to the BelongsToAssociation.

jhannus:
@pezra FYI I have another PR for compression which you can find here: https://github.com/cequel/cequel/pull/304. 

I'm working on specs for another local branch which adds a timestamp to batches. I was hoping to get some early feedback for it if you wouldn't mind. I noticed there is support for an optional `USING TIMESTAMP` for update statements but its a single statement only. I'm looking to add `USING TIMESTAMP` to a whole batch. This enables batch statement order execution provided each statement uses the same partition key. Very useful.

See: http://docs.datastax.com/en/cql/3.1/cql/cql_reference/batch_r.html?scroll=reference_ds_djf_xdd_xj__using-a-timestamp

pezra:
@jhannus, please remove the `.idea/*` files.

lib/cequel/record/belongs_to_association.rb
```diff
@@ -17,6 +17,8 @@ class BelongsToAssociation
       attr_reader :name
       # @return [String] name of parent class
       attr_reader :association_class_name
+      # @return [String] name of parent class
```
pezra: Docs need to be updated.

lib/cequel/record/belongs_to_association.rb
```diff
@@ -17,6 +17,8 @@ class BelongsToAssociation
...
+      attr_reader :foreign_key
```
pezra: Given that there can be multiple foreign key fields I think this option and attr should be plural.

spec/examples/record/associations_spec.rb
```diff
@@ -186,6 +186,51 @@
...
+      context 'with multiple foreign_keys' do
+        model :Foo do
+          key :account_id, :uuid, auto: true
+          key :parent_id, :uuid, auto: true
+          column :name, :text
+          has_many :children, class_name: 'Child'
+        end
+
+        model :Bar do
+          belongs_to :foo, partition: true, foreign_key: [:account_id, :parent_id]
```
pezra: Could we make the option plural too? Is that better?

PS: sorry about the delayed feedback. I'll merge this once we settle this last issue. 😄

jhannus: I'm impartial to plural/singular here. IMHO since a foreign key, or in this case partition key, can be compound... singular still describes it well. So, for instance, I don't think you would call a partition key "partition keys" when they're compound?

jhannus: @pezra either way please let me know and I'll update accordingly. And thanks again for all the feedback!

pezra: i see your point... neither name seem entirely clear to me. But! either is better than none at all. :smile:

pezra:
merged! Thanks for the improvement.

pezra:
released in 1.10.0

jhannus:
👍

PR #288, opened 2016-03-21
NewRelic instrumentation

shaqonline:
Instrumentation thru `NewRelic::Agent::Datastores` to make Cassandra visible as database.
Statistics are grouped by statement type + table name.
CQL visible only for slow queries.

Fixed config param name. (`newrelic` vs `newrelic_enabled`)

### DIFF
--- a/lib/cequel/metal/new_relic_instrumentation.rb
+++ b/lib/cequel/metal/new_relic_instrumentation.rb
@@ -15,7 +15,7 @@ module Cequel
 
       define_method :execute_with_consistency_with_newrelic do |statement, bind_vars, consistency|
         callback = Proc.new do |result, scoped_metric, elapsed|
-          NewRelic::Agent::Datastores.notice_sql(statement, scoped_metric, elapsed)
+          NewRelic::Agent::Datastores.notice_statement(statement, elapsed)
         end
 
         statement_words = statement.split
--- a/lib/cequel/metal/new_relic_instrumentation.rb
+++ b/lib/cequel/metal/new_relic_instrumentation.rb
@@ -19,16 +19,16 @@ module Cequel
         end
 
         statement_words = statement.split
-        operation = statement_words.first
+        operation = statement_words.first.downcase
         table = nil
         case operation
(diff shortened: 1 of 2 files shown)


pezra:
@shaqonline, if you rebase against master the specs should pass. I'd be delighted to have this functionality once the tests are green.

shaqonline:
@pezra one test is still red

pezra:
That is weird. It was some sort of order/eventual consistency issue, i think. I restarted the test which will likely solve this particular instance of the problem. I'll keep my eye out for that issue and if it shows up again i'll hunt it down for reals.

pezra:
@shaqonline, rather than injecting the instrumentation directly into the keyspace, how about making instances of `Keyspace` [observable](http://ruby-doc.org/stdlib-2.3.0/libdoc/observer/rdoc/Observable.html)? Keyspace could call `#changed` immediately after the query execution and pass the query and elapsed time to all observers. That would provide a generic way to logging, app perf tracking, etc to gain access to this information.

shaqonline:
@pezra, see how newrelic_rpm works for [ActiveRecord](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/active_record_helper.rb#L20) and [Redis](https://github.com/newrelic/rpm/blob/master/lib/new_relic/agent/instrumentation/redis.rb#L28). 

How about [Notifications](http://api.rubyonrails.org/classes/ActiveSupport/Notifications.html)?

pezra:
Using notifications would be a good generic solution.

Regarding how it is done with active record,  etc, is that out of necessity or convenience?

shaqonline:
@pezra necessity if we want to use NewRelic official method, [NewRelic::Agent::Datastores.wrap](https://docs.newrelic.com/docs/agents/ruby-agent/frameworks/third-party-instrumentation#wrap) block. But we can try to create our implementation of that and just report metrics without yield'ing.

pezra:
Merged! Thanks for the improvement

Issue #291, opened 2016-03-26
Unable to find class for belongs_to association


pranny:
I am creating a standalone ruby gem that has models in it. It is a simple gem that provides Blog/Post example from the readme. The dir structure goes like

```bash
- blogger.gemspec
- lib/
    - blogger.rb
    - blogger/
        - models.rb
        - models/
        	  - blog.rb
        	  - post.rb
        
```
The contents are as follows:

**blog.rb**

```ruby
require 'cequel'

module Blogger::Core
  class Blog
    include Cequel::Record
    key :id, :uuid
    has_many :posts
  end
end

```

**post.rb**

```ruby
require 'cequel'
require 'blogger/models/blog'

module Blogger::Core
  class Post
    include Cequel::Record
    belongs_to :blog
    key :id, :bigint
    column :created_at, :timestamp
  end
end

```

Once I install this gem and require it in irb, I get the error

```ruby

require 'blogger'
NameError: uninitialized constant Blog
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/activesupport-4.2.4/lib/active_support/inflector/methods.rb:261:in `const_get'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/activesupport-4.2.4/lib/active_support/inflector/methods.rb:261:in `block in constantize'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/activesupport-4.2.4/lib/active_support/inflector/methods.rb:259:in `each'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/activesupport-4.2.4/lib/active_support/inflector/methods.rb:259:in `inject'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/activesupport-4.2.4/lib/active_support/inflector/methods.rb:259:in `constantize'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/activesupport-4.2.4/lib/active_support/core_ext/string/inflections.rb:66:in `constantize'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/cequel-1.7.0/lib/cequel/record/belongs_to_association.rb:45:in `association_class'
	from /Users/u1/.rvm/gems/ruby-2.1.5/gems/cequel-1.7.0/lib/cequel/record/associations.rb:117:in `belongs_to'
[...]

PR #292, opened 2016-03-26
Adding example for namespaced models

pranny:
Referencing to this [issue](https://github.com/cequel/cequel/issues/291) and the [fix](https://github.com/cequel/cequel/pull/98). It is good to have it in the doc.


pezra:
Merged. Thanks!

PR #295, opened 2016-04-06
Adapt to Cassandra version to avoid CASSANDRA-8733

lesliev:
I've rebased this PR on master, the previous PR was here:
https://github.com/cequel/cequel/pull/252

This Cassandra bug causes list updates to reverse the order of the
array sent, which was being worked around in List#unshift by
reversing the array that was sent.

This change detects the version of Cassandra and only reverses
when it's a known buggy version. Also, the tests detect the
version and expect a list_prepend operation to produce reversed
lists in buggy versions.

See: https://issues.apache.org/jira/browse/CASSANDRA-8733
The bug was fixed in 2.1.3 and 2.0.13.

This code depends on the .cassandra-versions file to compile
a list of buggy versions, but doesn't add any other dependencies.

The following three tests now pass:
 Cequel::Record::List#unshift should persist unloaded unshift
 Cequel::Record::List#unshift should atomically unshift
 Cequel::Metal::DataSet#list_prepend should prepend multiple elements to list column

Tested on the following versions of Cassandra:
 2.0.4
 2.0.12
 2.0.13
 2.0.16
 2.1.2
 2.1.3
 2.1.8
 2.2.1


lesliev:
I've rebased on master so I hope this can be merged now. I've not rerun the tests because I don't think I have the VM's any more. Please let me know if that is required.

pezra:
That is a crazy bug. I have seen this issue but never looked into it.

What are the implications if the cluster is heterogeneous wrt versions? Eg, 1 server in the cluster has been upgraded to a version w/o the list reversal issue, but the rest of cluster has not? Which version matters? The server we are talking to or the owner of the token range for the record being written?

Also, i wonder how common version heterogeneous clusters are in the wild?

pezra:
@lesliev, any chance you could rebase this again?

lesliev:
Ok, I merged master and there were no conflicts. I've not run the tests again because I no longer have the Virtualbox images. 
Still not keen on merging this @outoftime? You seemed to be ok with it here: https://github.com/cequel/cequel/pull/252

pezra:
@outoftime no longer does much work on this project. I have taken over primary maintainership. I am definitely interested in merging this.

pezra:
i confirmed that the tests pass on cassandra 2.1 and 2.2.

pezra:
Thanks for the improvement, @lesliev.

lesliev:
Oh great, thanks for stepping up @pezra! About heterogeneous clusters, I think you are likely to get a version number reply from the first server to reply, but that's just a guess. I think such clusters would be rare since you'd only run a mixed cluster during an upgrade (or a fit of crazy). And since the bug here means you'd get the same operations causing different data on different versions, it would be especially crazy to straddle those versions for any length of time.

PR #296, opened 2016-04-08
Remove rubocop

pezra:


outoftime:
But consistent style!

pezra:
I'm open to the community's opinion on this one.

In my view rubocop provides little value and significant annoyance. I'd rather the code be clear, concise and expressive than consistent (though consistency is good where it doesn't detract from those other characteristics). I believe these values are better enforce by humans than machines and PR comments are a great vehicle for doing just that.

outoftime:
@pezra as you see fit! I think computers are the *perfect* entities to nitpick about consistent code style, but that’s just me : )

PR #300, opened 2016-05-13
switch instrumentation from execute_with_consistency to execute_with_options

bettse:
I hadn't ever gotten the new relic instrumentation to work, but I attributed it to configuration issues.  After diving in, I found that the changes I made a few months ago to support pagination may have broken the usefulness of instrumenting `execute_with_consistency` since it is now only used when doing batch queries.

Since `execute_with_consistency` calls `execute_with_options`, I thought it would make sense to change the instrumented method.  Trying it out locally, I was able to start seeing metrics and transaction segments for Cequel, as I never had before.


pezra:
Merged! Thanks for correcting the instrumentation.

pezra:
Released in 1.9.1

PR #304, opened 2016-06-14
Added client_compression setting and pass it to driver

jhannus:
Adds cassandra client compression to Cequel gem. 

This PR adds a new setting to the config to allow compression. It basically just passes it along to the Cassandra driver. 

See: http://datastax.github.io/ruby-driver/features/#compression for more info.

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -32,6 +32,8 @@ module Cequel
       attr_reader :credentials
       # @return [Hash] SSL Configuration options
       attr_reader :ssl_config
+      # @return [Symbol] The client compression option
+      attr_reader :client_compression
 
       #
       # @!method write(statement, *bind_vars)
@@ -139,6 +141,7 @@ module Cequel
 
         @name = configuration[:keyspace]
         @default_consistency = configuration[:default_consistency].try(:to_sym)
+        @client_compression = configuration[:client_compression].try(:to_sym)
 
         # reset the connections
         clear_active_connections!
@@ -284,6 +287,7 @@ module Cequel
         {hosts: hosts, port: port}.tap do |options|
           options.merge!(credentials) if credentials
           options.merge!(ssl_config) if ssl_config
+          options.merge!(compression: client_compression) if client_compression
         end
       end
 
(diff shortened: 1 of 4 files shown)


lib/cequel/metal/keyspace.rb
```diff
@@ -245,6 +248,13 @@ def default_consistency
         @default_consistency || :quorum
       end
 
+      #
+      # @return [Symbol] the client compression setting
+      #
+      def client_compression
+        @client_compression
+      end
```
pezra: Do we need this in addition to the attr reader defined on line 36?

spec/examples/metal/keyspace_spec.rb
```diff
@@ -104,6 +104,27 @@
...
+      Cequel.connect host: Cequel::SpecSupport::Helpers.host,
+          port: Cequel::SpecSupport::Helpers.port,
+          client_compression: client_compression
+    end
+    it "client compression settings get extracted correctly for sending to cluster" do
+      expect(connect.client_compression).to eq client_compression
+    end
+    it 'properly passes compression seetting to the driver' do
+      Cassandra.expects(:cluster).with(has_entries(compression: :lz4)).returns(client_mock)
+      expect(connect.client).to eq true
```
pezra: Is there anyway to verify this without a mock? Does the client object expose a compression attr, for example? I'd rather not use mocking if there is any other option since they basically just test that the implementation hasn't been changed.

jhannus:
Thanks the comments I'll address these later today after lunch :+1:

lib/cequel/metal/keyspace.rb
jhannus: To be honest, I don't think so. I added this to follow convention in keyspace.rb where other configuration options did the same.

spec/examples/metal/keyspace_spec.rb
jhannus: I can't seem to find a getter for the settings passed to the driver. IMHO the mock's expectation here verifies the the cequel gem properly passes the setting to the driver. I'm open changing this especially since no other tests uses mocks. But without a getter I'm not exactly sure how to verify this.

jhannus:
@pezra I removed the mock all together since it was causing some conflicts in the rest of the test suite. I also removed the redundant getter for `client_compression`. Thanks for your feedback!

pezra:

jhannus:
:+1:

pezra:
released in 1.10.0

PR #307, opened 2016-07-15
update link target urls in README

atxviking:


pezra:
Great update. Thanks!

PR #311, opened 2016-07-22
Support blob values

dividedmind:
Blob values weren't supported correctly due to cequel doing its own statement sanitization (skipping hex encoding). Since cassandra-driver supports statements with arguments now, use that instead.


pezra:
@dividedmind, we should definitely delegate this behavior to the driver. 👍

Did you mean to close this PR?

dividedmind:
Yes, because it broke a couple of things. Still working on it.
24 lip 2016 3:47 PM "Peter Williams" <notifications@github.com> napisał(a):

PR #313, opened 2016-07-28
limit is not applied to the collection before it is counted

perobertson:
Ive been working on upgrading to cassandra 3 and came across a difference in the way `LIMIT` functions. In older versions of cassandra it seems that the limit is applied to the data set before it is counted. In later versions it applies the limit to the number of rows that were returned, which is always 1.

I havent tested this out on the different vagrant images to see when this behaviour change occurred, but it is somewhere between `Cassandra 2.1.13.1131` and `Cassandra 3.0.7.1159`

/cc @orenmazor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Unreleased
 
 * drop support for JRuby 1.7
+* Dropped support for count, length, and size as it results in unbounded run times and memory usage
+  ([PR 313](https://github.com/cequel/cequel/pull/313))
 
 ## 1.10.0
 
--- a/README.md
+++ b/README.md
@@ -534,6 +534,14 @@ essentially the same thing: both simply persist the given column data at the
 given key(s). So, you may think you are creating a new record, but in fact
 you're overwriting data at an existing record:
 
+#### Counting ####
+
+Counting is not the same as in a RDB, as it can have a much longer runtime and
+can put unexpected load on your cluster. As a result Cequel does not support
+this feature. It is still possible to execute raw cql to get the counts, should
+you require this functionality.
+`MyModel.connection.execute('select count(*) from table_name;').first['count']`
+
 ``` ruby
 # I'm just creating a blog here.
 blog1 = Blog.create!(
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -615,7 +615,12 @@ module Cequel
       # @return [Fixnum] the number of rows in this data set
       #
       def count
-        execute_cql(*count_cql).first['count']
+        actual_count = execute_cql(*count_cql).first['count']
+        if row_limit
+          [actual_count, row_limit].min
+        else
+          actual_count
+        end
       end
 
       #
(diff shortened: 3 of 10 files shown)


pezra:
I wonder if this change makes `#count` dangerous and useless enough to warrant removing it altogether? Any nontrivial size table is going to be too slow to count this way and the removal of the ability to limit the number of rows counted means the runtime of such an operation is unbounded. The loss of the runtime bounding warrants a major version bump so if we want to drop that functionality now is the time.

orenmazor:
@pezra valid suggestion. I guess it comes down to how opinionated we want Cequel to be? ActiveRecord lets you do some incredibly stupid things. Redis drivers *still* let you do a `keys *`, even tho pretty much every project I've been part of monkey patches that command out

perobertson:
I think it makes more sense to drop the `count` functionality now than to carry it forward. If we do drop it, the readme will should get updated with how you can run a raw cql query to retrieve the count for the people who rely on that function. However, I am not actually sure how you can run a raw query from cequel from a project using this gem.

pezra:
I vote for being opinionated and dropping `#count`. There is no need to provide functionality that we know doesn't really work as advertised.

pezra:
Raw CQL can be executed like `MyModel.connection.execute("select count(*) from my_models")`.

pezra:
Also, it looks like rubocop doesn't play nice with jruby 1.7 anymore. We should drop support for jruby 1.7 since it is the ruby 1.9 compatible jruby and ruby 1.9 has not been maintained for a long time.

perobertson:
Ive been thinking about this and I think it makes sense to remove the `.count` code, but place it in a module that people can include on their record. That way people would have to opt into potentially bad queries.

pezra:
I'm not sure I like having the count function at all. I recently ran into a situation where some rarely used code in our system used count and it crashed the Cassandra servers. All of them. That's a pretty serious drawback. I think updating the read me to explain why one probably doesn't want to count would be better than providing functionality that's known to be dangerous.

orenmazor:
Our life is spent dealing with count misuse. :+1: for killing + documenting
why.

perobertson:
I just updated this PR with a commit that removes the counting functionality from cequel. I found that I had to raise an exception rather than not defining the method because RecordSet would forward the method call to enumerable which would load up the records into memory and then count the size of the collection. `Post['cassandra'].count` would cause the loading of the records where `Post.count` would not. Now in all cases it will raise `Cequel::Record::DangerousQueryError`

orenmazor:
I would've gone with DangerZoneException but maybe that's just me :)

pezra:
@perobertson, looks good. I am working on getting the builds working again. Hopefully that won't take long and we can get this merged soon.

pezra:
@perobertson, if you rebase onto master the tests should pass.

pezra:
Can you rebase one more time. I promise i'll merge it this time. 😄

perobertson:
rebased but it looks like travis had a hiccup when booting cassandra for one of the builds

PR #314, opened 2016-07-28
allow for creation of cassandra 3 vms

perobertson:
I updated the vagrant file to allow for the creation of vms that will work with cassandra 3. This will come in handy when upgrading support.


pezra:
Thanks for the improvement.

PR #315, opened 2016-08-25
Support Rails 5

emn178:
Hi,

I upgraded to support Rails 5.

There are 3 failed unit tests in origin master. I'm not sure the correct usage so that I did not fix them in this commit.

```
Failures:

  1) Cequel::Metal::DataSet#list_prepend should prepend multiple elements to list column
     Failure/Error:
       expect(cequel[:posts].where(row_keys).first[:categories]).to eq(
         ['Partition Tolerance', 'Scalability', 'Big Data', 'Cassandra']
       )

       expected: ["Partition Tolerance", "Scalability", "Big Data", "Cassandra"]
            got: ["Scalability", "Partition Tolerance", "Big Data", "Cassandra"]

       (compared using ==)
     # ./spec/examples/metal/data_set_spec.rb:199:in `block (3 levels) in <top (required)>'

  2) Cequel::Record::List#unshift should atomically unshift
     Failure/Error: expect(subject[:tags]).to eq(%w(minustwo minusone zero one two))

       expected: ["minustwo", "minusone", "zero", "one", "two"]
            got: ["minusone", "minustwo", "zero", "one", "two"]

       (compared using ==)
     # ./spec/examples/record/list_spec.rb:429:in `block (3 levels) in <top (required)>'

  3) Cequel::Record::List#unshift should persist unloaded unshift
     Failure/Error: expect(subject[:tags]).to eq(%w(minustwo minusone one two))

       expected: ["minustwo", "minusone", "one", "two"]
            got: ["minusone", "minustwo", "one", "two"]

       (compared using ==)
     # ./spec/examples/record/list_spec.rb:446:in `block (3 levels) in <top (required)>'

Finished in 2 minutes 51.2 seconds (files took 1.95 seconds to load)
814 examples, 3 failures, 1 pending

Failed examples:

rspec ./spec/examples/metal/data_set_spec.rb:194 # Cequel::Metal::DataSet#list_prepend should prepend multiple elements to list column
rspec ./spec/examples/record/list_spec.rb:425 # Cequel::Record::List#unshift should atomically unshift
rspec ./spec/examples/record/list_spec.rb:443 # Cequel::Record::List#unshift should persist unloaded unshift
```


pezra:
Thanks for the PR. I'm really busy this week so it will be a few days before I can look into this.

pezra:
@emn178, it looks like the CI builds are failing because of the new dependency on `activemodel-serializers-xml`. I don't see this being used in Cequel (nor would i expected to). Could we just remove that from the `Gemfile`?

emn178:
Hi, 
Rails 5 removed built-in `activemodel-serializers-xml`. We need to add this by ourself. I think it can be removed from gemspec for passing CI. Users still need to add this into Gemfile, and I wrote in README. I will update later.

pezra:
@emn178, is the xml serializer a requirement for your app or do you think it is required for every rails 5 app and uses cequel?

emn178:
It can not pass unit tests without `activemodel-serializers-xml`. It's required for rails 5 and cequel.

pezra:
@emn178, we can [exclude the 2.1.5/rails build](https://docs.travis-ci.com/user/customizing-the-build#Excluding-Jobs) and upgrade the ruby 2.2.0 builds to 2.2.2.

emn178:
ok, but it still has problems in jruby. I will check it later.

emn178:
jruby has bug in 9.0.0.0. of prepend. I upgraded to 9.1.5.0 and prepend works. But there is the other bug. I can not find a workable version to fix it. :(

pezra:
I suppose we could drop support for jruby (with an appropriate major
version bump). I suspect that the audience for rails 5 is bigger than for
jruby. I don't love removing functionality but that might be the least bad
option. I tried to upgrade to jruby 9.1.5.0 recently myself but gave up due
to the same, i assume, bug you encountered.

PR #316, opened 2016-08-30
Port gem to Cassandra 3

hydrogen18:
I upgraded this gem to use Cassandra 3. It is not reverse compatible with prior versions of Cassandra, so I raised the gem major version.

The tests passed for me against 3.7.

This code is rather interested in the the metadata around table schemas. That was changed substantially between 2 & 3. I reverse engineering the python & ruby drivers for Cassandra3 as needed to figure out how to query all of that stuff. I tried to use the constants defined by the ruby driver gem where appropriate. Compact tables are really annoying, and apparently have no use in Cassandra3 according to this:

http://www.datastax.com/2015/12/storage-engine-30

It might be worthwhile to just drop compact storage support in the future.

I also added an option to allow the passing of arbitrary options to the underlying Cassandra gem. I need to be able to reconfigure the node selection strategy as well as other parameters. I don't think this hurts anything as users clearly are on their own if they start reconfiguring the underlying client.

Let me know if you need anything else to accept this MR.

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -34,6 +34,8 @@ module Cequel
       attr_reader :ssl_config
       # @return [Symbol] The client compression option
       attr_reader :client_compression
+      # @return [Hash] A hash of configuration options passed to Cassandra.cluster, with absolute precedence 
+      attr_reader :cluster_options
 
       #
       # @!method write(statement, *bind_vars)
@@ -138,6 +140,7 @@ module Cequel
         @max_retries  = extract_max_retries(configuration)
         @retry_delay  = extract_retry_delay(configuration)
         @ssl_config = extract_ssl_config(configuration)
+        @cluster_options = configuration[:cluster_options]
 
         @name = configuration[:keyspace]
         @default_consistency = configuration[:default_consistency].try(:to_sym)
@@ -284,6 +287,7 @@ module Cequel
           options.merge!(credentials) if credentials
           options.merge!(ssl_config) if ssl_config
           options.merge!(compression: client_compression) if client_compression
+          options.merge!(cluster_options) if cluster_options
         end
       end
 
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -250,12 +250,8 @@ module Cequel
 
       # @return [Boolean] true if the keyspace exists
       def exists?
-        statement = <<-CQL
-          SELECT keyspace_name
-          FROM system.schema_keyspaces
-          WHERE keyspace_name = ?
-        CQL
-
+        statement = Cassandra::Cluster::Schema::Fetchers::V3_0_x::SELECT_KEYSPACE 
+        
         log('CQL', statement, [name]) do
           client_without_keyspace.execute(sanitize(statement, [name])).any?
         end
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -301,7 +301,7 @@ module Cequel
 
       def extract_hosts_and_port(configuration)
         hosts, ports = [], Set[]
-        ports << configuration[:port] if configuration.key?(:port)
+        ports << configuration[:port] if configuration[:port].present?
         host_or_hosts =
           configuration.fetch(:host, configuration.fetch(:hosts, '127.0.0.1'))
         Array.wrap(host_or_hosts).each do |host_port|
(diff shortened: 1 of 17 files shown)


lib/cequel/version.rb
```diff
@@ -1,5 +1,5 @@
 # -*- encoding : utf-8 -*-
 module Cequel
   # The current version of the library
-  VERSION = '1.10.0'
+  VERSION = '2.s0.0'
```
pezra: typo 😄

spec/examples/record/record_set_spec.rb
```diff
@@ -674,7 +674,9 @@ def self.latest(count)
     let(:records) { blogs }
 
     it 'should return the number of records requested' do
-      expect(Blog.limit(2).length).to be(2)
+      # The method length is aliased to count, which does not work as expected 
+      # in Cassandra 2.x
+      expect(Blog.limit(2).find_each.to_a.length).to eq(2)
```
pezra: Is the `find_each` really necessary?

lib/cequel/version.rb
hydrogen18: Thanks, will address

lib/cequel/metal/keyspace.rb
```diff
@@ -311,16 +311,19 @@ def extract_hosts_and_port(configuration)
...
         if ports.size > 1
           fail ArgumentError, "All Cassandra nodes must listen on the same " \
                "port; specified multiple ports #{ports.join(', ')}"
+        elsif ports.size != 1
+          fail ArgumentError, "Port for cassandra nodes not specified"
         end
+        
 
-        [hosts, ports.first || 9042]
+        [hosts, Integer(ports.first) || 9042]
```
perobertson: You should probably use `Integer(ports.first || 9042)` instead because integer will fail to parse nil if no ports are supplied

perobertson:
We should maintain some form of backwards compatibility with Cassandra 2 so that people can use a specific version of the gem when upgrading their cluster from v2 to v3. Im also open to suggestions on how we can entirely drop support for Cassandra 2 and do a cluster upgrade at the same time. When I was playing around with upgrading this gem I found I could use the new version of the driver on all versions. I only ran into issues when running migrations.

spec/examples/record/record_set_spec.rb
hydrogen18: This is just a refactor of code I found elsewhere in the specs.

hydrogen18:
I pushed up my changes to address the code comments. Can anyone point me to docs. to reconfigure travis to run w/ cassandra 3? it appears it is running but with cassandra 2

pezra:
i'd love to have support for c* 3 but i am not willing to give up support for 2.1 and 2.2 just yet. Do you think it possible to support both?

pezra:
Re your travis question, i don't think there is any way to change the c* version in their containerized builds. We could switch to VM based builds and [use docker to start which ever version we want](https://docs.travis-ci.com/user/docker/).  Combined with the `env` part of the [build matrix](https://docs.travis-ci.com/user/customizing-the-build#Build-Matrix) we could even build against multiple versions of c*.

hydrogen18:
It is probably possible to somehow abstract out most of the operations to support Cassandra 2 & 3 on the same branch. You would need some configuration flag to indicate what version the cluster is that is being connected to.

Given that the gem has no such distinctions presently, that seems like a pretty involved refactor that I don't think I have the time to take on. This would be easier if there was documentation from DataStax about what changed and what things mean, but I couldn't find anything. Hence, I just reverse engineered their driver to figure out how things are expected to work.

Why not just make the `1.x` versions of the gem be the only ones supporting Cassandra 2?

As far as the Travis changes go, I was thinking I just needed to change a version number somewhere. I'll leave it up to you to figure that out.

pezra:
> Why not just make the 1.x versions of the gem be the only ones supporting Cassandra 2?

I don't want to maintain to increasingly divergent version of this library.  Also, i believe in semantic versioning. Creating 2.x branch also implies a commitment to never making breaking changes in the branch that supports c* 2.x. I am not willing to make such a commitment. Consider #315 where we may (temporarily) drop support for jruby.

Fwiw, my day job uses c* 2.2 so i have a vested interest in continuing to support that for a bit longer.

hydrogen18:
I'll need to do some more investigation before I can decide if the time is available to support Cassandra 2 & 3 on the same branch.

One issue is the only reason I am working on this is to get the last remaining application off Cassandra 2. So it'd be almost impossible for me to dogfood this as no one I work with is going to get excited about Cassandra 2.

Hypothetically, would you be OK with the gem starting in Cassandra 2 mode and then the user doing something like this 

```
Cequel.cassandra_3!
```

This would just shift the gem into Cassandra 3 mode. We wouldn't support accessing one cluster that is Cassandra 2 and another that is Cassandra 3 from the same Ruby application.

pezra:
I definitely be OK with that. I think supporting only one version of c* per ruby application is very reasonable. I'd prefer that cequel figure out which version automatically when it connects to the cluster but that is not a requirement.

perobertson:
Inside `Cequel::Metal::Keyspace` I was able to do this to determine what version we are connecting to `cluster.hosts.first.release_version`. The issue there was that our migration plan was going to cycle in Cassandra 3 nodes to replace the old ones. You could probably scan the release version of each node to see if any of them were a 2.x release.

pezra:
@perobertson, did adding c* 3 nodes to a 2.x cluster work? It would definitely be nice to be able to upgrade like that without taking down applications based on cequel.

perobertson:
We have not actually started the upgrade process yet, but we have talked to other people who run larger clusters and have confirmed that this is the approach we should take. Also, the [DSE Guide](https://docs.datastax.com/en/latest-upgrade/upgrade/datastax_enterprise/upgdDseMinor.html) recommends doing it this way and in a specific order. I think the guide will still work for people not running the DSE distribution.

pezra:
closing in favor of #324

PR #317, opened 2016-09-04
drop index if it exists

perobertson:
Cequel gets confused of the schema state when an index is dropped outside of a cequel migration. We should only drop the index when it exists.


pezra:
👍 once you rebase. (maybe after i merge you count PR?)

perobertson:
rebased and either PR first works for me. It doesnt take much to get them rebased.

pezra:
rebase on more time please

perobertson:
rebased

Issue #206, opened 2014-12-23
cequel/record/finders.rb:93: duplicated argument name


swistaczek:
Hello,
I have recently tried **cequel 1.5.0**. First please take a look at my Cequel models:

``` ruby
class Account
  include Cequel::Record
  key :name, :text
  column :description, :text
  has_many :api_keys
end

class ApiKey
  include Cequel::Record
  belongs_to :account
  key :id, :timeuuid, auto: true
  column :secret, :text
end
```

After running `rake cequel:migrate`I recived following error:

``` ruby
SyntaxError: /Users/ernest.bursa/.rvm/gems/jruby-1.7.17@meta_store/bundler/gems/cequel-6935b245fef4/lib/cequel/record/finders.rb:93: duplicated argument name
          def find_by_name_and_name(name, name)
                                              ^
/Users/ernest.bursa/.rvm/gems/jruby-1.7.17@meta_store/bundler/gems/cequel-6935b245fef4/lib/cequel/record/finders.rb:90:in `def_finder'
/Users/ernest.bursa/.rvm/gems/jruby-1.7.17@meta_store/bundler/gems/cequel-6935b245fef4/lib/cequel/record/finders.rb:75:in `def_key_finders'
/Users/ernest.bursa/.rvm/gems/jruby-1.7.17@meta_store/bundler/gems/cequel-6935b245fef4/lib/cequel/record/finders.rb:61:in `key'
/Users/ernest.bursa/.rvm/gems/jruby-1.7.17@meta_store/bundler/gems/cequel-6935b245fef4/lib/cequel/record/dirty.rb:22:in `key'
/Users/ernest.bursa/.rvm/gems/jruby-1.7.17@meta_store/bundler/gems/cequel-6935b245fef4/lib/cequel/record/timestamps.rb:36:in `key'
```

After small research I found that this error is rised due to #key_columns:

``` ruby
key_columns => [#<Cequel::Schema::PartitionKey: name text>, #<Cequel::Schema::ClusteringColumn: name text>]

method_suffix => "name_and_name
```

I have no expirence with Cequel and its internals, I would be very grateful if you could share any idea how to solve that issue.

wflanagan:
So, this is "standard" behavior in rspec-preloader, which we use in our development workflow. It reloads any files that change. And, the consequence, we we get this error and can't use rspec-preloader.

pezra:
[...]

PR #319, opened 2016-09-11
Fixes Issue #206 duplicated argument name

wflanagan:
as well as minor changes for compatibility with non-Rails environments, including: 

1) Supports CEQUEL_MODELS_PATH env var so you can point to a destination besides app/models for loading of files.  Does not handle namespaced models in this folder yet, but its easy to do from here. I might do that in a later request.  
2) Downgrade to Rails 3.2 compatibility to be in compliance with documentation, which states that it supports > 3.1.
3) Adds a check to ensure that failed rspec tests on one run don't mess up the runs of subsequent tests.  

How the fix is implemented:
Changes to the duplicated argument name were fixed by changing accessors from Arrays to Hashes, and then using the hash.values to set the arrays.  I also implemented minor checks to ensure that some weird edge case won't create further problems. 
I also broke out some fatter methods in the table_writer.rb to make it easier to make changes in the future.   

Notes: 
It was several days in before I realized that the CEQUEL_TEST_HOST env var was available. This should be made more prominent in the documentation, as its required for Docker compatibility.


lib/cequel/schema/table.rb
```diff
@@ -157,11 +157,26 @@ def data_columns
...
+          index_name = options[:index]
+          index_name = :"#{@name}_#{name}_idx" if index_name == true
+          DataColumn.new(name, type(type), index_name)
+            .tap do |column|
+              data_columns_hash[name] = add_column(column)
+            end
+        else
+          # return the data column like we would if we built it
+          data_columns_hash[name]
+        end
```
pezra: How would you feel about raising an exception if the column being defined didn't match the previously defined column? That might prevent some subtle errors.

lib/cequel/record/tasks.rb
```diff
@@ -63,12 +63,15 @@ def drop!
     puts "Dropped keyspace #{Cequel::Record.connection.name}"
   end
 
+  def models_dir_path
+    ENV['CEQUEL_MODELS_PATH'] || "#{File.expand_path('app/models', project_root)}/"
```
pezra: `ENV.fetch('CEQUEL_MODELS_PATH', project_root + "app/models")`? We'd have to ensure that `project_root` was a `Pathname` but that would be a Good Thing.

lib/cequel/schema/table.rb
```diff
@@ -1,5 +1,6 @@
 # -*- encoding : utf-8 -*-
 require 'stringio'
+require 'set'
```
pezra: Is the `Set` class used?

lib/cequel/schema/table.rb
```diff
@@ -59,17 +48,50 @@ def initialize(name)
       # @see #add_partition_key
       #
       def add_key(name, type, clustering_order = nil)
-        if @partition_key_columns.empty?
-          unless clustering_order.nil?
-            fail ArgumentError,
-                 "Can't set clustering order for partition key #{name}"
+        if valid_key?(name)
+          if partition_key_columns.empty?
```
pezra: 👍 for using reader methods!

pezra:
On the whole the changes in this PR look really good.

There is some sort of DNS issue with rubygems.org so i cannot restart that one failed build at the moment but i will tomorrow. I am pretty sure the it will pass (our tests are a little flakey).

lib/cequel/record/associations.rb
```diff
@@ -100,26 +100,30 @@ module ClassMethods
...
-          if table_schema.key_columns.any?
-            fail InvalidRecordConfiguration,
-                 "belongs_to association must be declared before declaring " \
-                 "key(s)"
-          end
+            # skip adding the belongs to if already defined on a reload
+            unless parent_association.association_class_name.downcase.to_sym == name.to_sym
+              fail InvalidRecordConfiguration,
+                   "Can't declare more than one belongs_to association"
+            end
```
wflanagan: Sure, i can do that. I will make the change in the next few days and push it.

lib/cequel/schema/table.rb
wflanagan: On this.. i saw it was used in the class.. in troubleshooting, i wanted to be sure that things were OK.. i will remove in the revision.

lib/cequel/record/tasks.rb
wflanagan: Not sure I get what you're saying on this one. This is operational for me using the CEQUEL_MODELS_PATH to point to an alternative location for the model files.  Help me understand and I'm happy to patch.

lib/cequel/schema/table.rb
wflanagan: Yeah, I thought about that a lot, and how that would work in the case of migrations. But, then, I decided that made my head hurt.. so I did it this way. I think though that it's a good thought, and I definitely thought it as well. I'll see what I can do on my next pull request.

lib/cequel/record/tasks.rb
pezra: I'm saying that i like `Hash#fetch` and `Pathname#+` better than `Hash#[]` and `File.expand_path`. 

It is a small thing, on which reasonable people can disagree, and is not necessary to merge this PR.

pezra:
Any chance i could get this change on 2.0? I'd love to have it.

wflanagan:
I'll see what I can do.. had to move off of Cassandra for a bit.. as soon
as I get back to it (hopefully a few days), I can push an update.

wflanagan:
Hey, what's the status here? Should I use this and try to upgrade, or did you make big changes on 2.0 that make this irrelevant?

lib/cequel/schema/table.rb
wflanagan: Ok.. I've upgraded to 2.0.3 and I'm implementing these changes.  This is done.

lib/cequel/schema/table.rb
wflanagan: Done.

lib/cequel/record/tasks.rb
wflanagan: I've done it, I think. :-) I set it in the "tell don't ask" style that checks for Rails::Railtie (much more reliable than Rails, as many people monkey patch the Rails module into existence in non-rails systems). If there, will use that folder. If not, will use the Dir.pwd as was used in the previous version. Then, I check the resulting pathname to see if it exists. If not, I raise a NoModelDirectoryError.

pezra:
@wflanagan, the changes for 2.0 are relatively minor, but still breaking, changes. Thanks for moving forward.

pezra:
Something weird is going on with the travis build. I am looking into it. I'll let you know once i know what is up.

wflanagan:
Let me know what they are, and I’ll try to fix.

WF

pezra:
@wflanagan, can you add a commit to this PR? Maybe add a blank line to the end of the README or something?

wflanagan:
Put a space in the Vagrantfile.  Let's see what happens.

pezra:
@wflanagan, just wanted you to know i haven't forgotten this. Hopefully #356 will get this branch to a green build.

wflanagan:
No worries. I’m trucking on my branch.. when it gets in, if it gets in,
I’ll swing to your branch.

WF

pezra:
@wflanagan, i don't know what is up with this branch exactly, but it fails at a much higher frequency than master, or any other branch. That indicates there is a real problem in the code. It is probably a race condition in the code, or order dependency in the tests given the intermittency of failure.

I do like the change and if we can get the build green i will merge it, but until then i cannot.

wflanagan:
Of course. Can you point me at a bit of the code that’s problematic? If so,
I’ll take a crack at fixing.

pezra:
I don't currently have any clue as to the root of the problem. The failures are intermittent and not always in the same tests but the failures all seem related to changing the schema.

wflanagan:
Hey, on this.. I"ve updated to Cequel 3.0 myself, changed libraries, etc. and effectively made this irrelevant. Closing.  I'll come back and contribute in other ways.. ;-)

Issue #306, opened 2016-06-29
Proposal for prepared statements


jhannus:
Hello Cequel Developers,

I'm hoping to start a discussion about the possibility of adding prepared statements to the cequel gem. I'm thinking that at a high-level prepared statements could be introduced (possibly in this order):

- The finder methods generated for each model by cequel. These seem like the best candidate and the lowest hanging fruit since they are already auto generated and are single use case driven.
- Associations `has_many` and `belongs_to` (for similar reasons as above)
- #save, #update, and #delete
- Custom queries i.e. `Model.select(:foo).where(:condition).prepared` where `prepared` tells cequel to lately bind a prepared statement?
- Usage in batches?

I'm thinking that maybe a configuration option would tell cequel to use prepared statements for the auto generated finders, associations, and save methods where the `prepared` method would always be available to invoke on relation objects.

Thoughts?

Thanks!
-j

pezra:
Awesome idea. We should be using prepared statements where possible.

I'd be inclined to have prepared statements be the only way finders and associations work. Are there downsides to using prepared statements that necessitate an option to revert to the dynamic statement behavior?

It seems like it might be easier to start with custom queries because it is more isolated.  However, what to do first is a decision best left to whoever decides to put together the first PR. :)

jhannus:
> Are there downsides to using prepared statements...
None that I can think of aside from some edge-case-like stuff. It just seems like some folks might want the option? Maybe this piece can be omitted from start. If needed it could be added later on.

> It seems like it might be easier to start with custom queries because it is more isolated
Good point. I could imagine a single mechanism is responsible for locating prepared statements before executing raw CQL so that finders, relations, and custom queries can technically be done in any order with that affordance.

I'm trying to get this planned for my next sprint. I'll let you know how it goes.

👍

pezra:
Sounds good. I look forward to seeing the PR. :smile:

jhannus:
tldr;
I have a basic working version of prepared statements for the above proposal including use within batches. The amount of refactoring is significant and the code is not ready. I'll explain below :)

Ok @pezra I have a _decent_ working version of this in my internal Github which I plan on cleaning up a little bit and moving to my public Github fork for early review. Before doing so I'd like to call out a few things because it's not finished enough for an open source contribution/merge but is enough for my apps use cases and testing.

[...]

PR #320, opened 2016-09-13
Statements refactor

jhannus:
Here are two separated commits of the statement refactor @pezra and I talked about. The type hinting was only applied to some of the writers. I'm not exactly sure I like the way the type hints are being passed around so maybe I could get some feedback/suggestions on it before continuing?

p.s. some tests may be failing ATM

Started from discussion: https://github.com/cequel/cequel/issues/306

### DIFF
--- a/lib/cequel/metal/statement.rb
+++ b/lib/cequel/metal/statement.rb
@@ -11,17 +11,29 @@ module Cequel
       # @return [Array] bind variables for CQL string
       attr_reader :bind_vars
       # @return [Array] cassandra type hints for bind variables
-      attr_accessor :type_hints
 
-      def initialize(cql='', bind_vars=[], type_hints=[])
-        @cql, @bind_vars, @type_hints = cql, bind_vars, type_hints
+      def initialize(cql_or_prepared='', bind_vars=[])
+        cql, prepared = *case cql_or_prepared
+                         when Cassandra::Statements::Prepared
+                           [cql_or_prepared.cql, cql_or_prepared]
+                         else
+                           [cql_or_prepared.to_s, nil]
+                         end
+
+        @cql, @prepared, @bind_vars = cql, prepared, bind_vars
       end
 
       #
       # @return [String] CQL statement
       #
-      def cql
+      def to_s
         @cql
+      end
+      alias_method :cql, :to_s
+
+      # @return [Cassandra::Statements::Prepared] prepared version of this statement
+      def prepare(keyspace)
+        @prepared ||= keyspace.client.prepare(cql)
       end
 
       #
--- a/lib/cequel/metal/statement.rb
+++ b/lib/cequel/metal/statement.rb
@@ -11,7 +11,7 @@ module Cequel
       # @return [Array] bind variables for CQL string
       attr_reader :bind_vars
       # @return [Array] cassandra type hints for bind variables
-      attr_reader :type_hints
+      attr_accessor :type_hints
 
       def initialize(cql='', bind_vars=[], type_hints=[])
         @cql, @bind_vars, @type_hints = cql, bind_vars, type_hints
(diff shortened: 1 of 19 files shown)


pezra:
I didn't mean to push your branch, @jhannus. I assumed the push would be blocked due to access control. I hope you don't mind.

jhannus:
No worries at all just possible failing specs lol. I'll take a look this coming morning +1

pezra:
If you don't mind i'll keep chipping away at the specs until they are all passing.

pezra:
I also rebased. In for a penny, in for a pound, as they say.

jhannus:
Yes please! I was hoping to find a better way to pass the type hints to the writers. So that's not finished. 

I'm thinking that instead of passing table and columns names to the writers we can pass a schema and figure out the names from there. Since the writers are responsible for assembling the query... We could then figure the queries column types hints from there. 

With writing to a single encapsulated statement object each query whether single or batch, prepared or not, it will contain exactly it's own parameters.

pezra:
@jhannus, let me know what you think. Type hints turned out be to really hard to get working. Once i switch to just preparing everything it all pretty much fell into place.

jhannus:
👍 @pezra thank you. Yeah the type hints where not fun at all but required to optionally prepare. In the fork I have hosted internally thats the way I went. I figured to contribute back though it should be optional possibly for folks who dynamically generate queries but... I'm cool with all prepared all the time if it cool with you 😄

jhannus:
I cherry picked the `PreparedStatementManager` I was talking about plus setters/getters and specs but its still unused on execute. I would remove the keyspace reference from the `Statement` instance (less object refs) and use `Cequel::Metal::Keyspace#prepared_statement` from `#execute` instead.

lib/cequel/metal/statement.rb
```diff
@@ -10,16 +10,30 @@ module Metal
...
-      def cql
-        @cql.join
+      def to_s
+        @cql
+      end
+      alias_method :cql, :to_s
+
+      # @return [Cassandra::Statements::Prepared] prepared version of this statement
+      def prepare(keyspace)
+        @prepared ||= keyspace.client.prepare(cql)
```
jhannus: I think my previous comment on this was canceled somehow but...
Memoizing the prepared CQL is good here but each `Cequel::Metal::Statement` object will ask the server for  a new prepared request even if this statement was already prepared once.

To mitigate this I cherry-picked my `PreparedStatementManager` which reuses prepared statements per `Cequel::Metal::Statement` instance. Right now the `Keyspace` only has the setter/getter for it. We'll need to use `Keyspace#prepare_statement` this during execute instead and remove `Cequel::Metal::Statement#prepare`

pezra:
@jhannus, it looks to me like the cassandra driver caches prepared statements and prevents duplicates (search <https://github.com/datastax/ruby-driver/blob/v2.1.7/lib/cassandra/cluster/client.rb> for `@prepared_statements[host][cql]`). Given that do we need the statement manager?

jhannus:
That would be great. When I read these lines I thought otherwise but I'm seeing other code where they are cached. https://github.com/datastax/ruby-driver/blob/v2.1.7/lib/cassandra/cluster/client.rb#L224-L236. I'll remove those commits.

jhannus:
I rebased to remove those commits 👍

pezra:
If you rebase i think this is ready to merge

jhannus:
OK all rebased to master.

pezra:
one more time. i merged another PR at the same time you did this.

pezra:
Looks like a few things need to change due to the upgrade of the cassandra driver gem.

jhannus:
OK, I got those changes in there as well. We'll see from the spec run if there are still issues with the upgrade.

pezra:
Rebased because i merged another PR.

pezra:
@jhannus, i rebased and, i think, fix the last remaining error. However that last rebase was a little weird. Could skim the current set of commits to ensure sure they make sense.

jhannus:
Sure thing...

jhannus:
Looks good 👍

pezra:
merged! thanks for the improvement and the stick-to-it-iveness to get it to completion.

PR #329, opened 2016-10-05
Don't include xml serializers when the gem is not available

pezra:


pezra:
released in version 2.0.1

PR #330, opened 2016-10-10
retry preparing statements when there is a cassandra error

perobertson:
I started discovering new Cassandra Errors being propagated up when updating to v2 of this gem. I traced it down to the fact that preparing the statement can raise most of the same errors as the execution. I also thought it was appropriate to switch to catching the errors specified in the docs. 

[prepare docs](http://docs.datastax.com/en/developer/ruby-driver/3.0/api/cassandra/session/#prepare-instance_method)
[execute docs](http://docs.datastax.com/en/developer/ruby-driver/3.0/api/cassandra/session/#execute-instance_method)

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -205,7 +205,7 @@ module Cequel
         retries = max_retries
         cql, options = *case statement
                         when Statement
-                          [client.prepare(statement.cql),
+                          [prepare_statement(statement),
                            {arguments: statement.bind_vars}.merge(options)]
                         when Cassandra::Statements::Batch
                           [statement, options]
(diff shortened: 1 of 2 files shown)


spec/examples/metal/keyspace_spec.rb
```diff
@@ -126,16 +126,70 @@
...
 
         expect { cequel.execute(statement) }.not_to raise_error
       end
+
+      it "reconnects to cassandra with a new client after execution failed" do
+        allow(cequel.client)
+          .to receive(:execute)
+               .with(->(s){ s.cql == statement},
+                     hash_including(:consistency => cequel.default_consistency))
+          .and_raise(Cassandra::Errors::NoHostsAvailable)
```
perobertson: This was actually supposed to raise `Cassandra::Errors::ExecutionError`

spec/examples/metal/keyspace_spec.rb
```diff
@@ -126,16 +126,70 @@
...
+    context "without a connection error" do
+      it "executes a CQL query" do
+        expect { cequel.prepare_statement(statement) }.not_to raise_error
+      end
+    end
+
+    context "with a connection error" do
+      it "reconnects to cassandra with a new client after no hosts could be reached" do
+        allow(cequel.client)
+          .to receive(:prepare_statement)
```
perobertson: the expected receives were supposed to be just `prepare`

pezra:
It makes completel sense that prepare would fail that way. Thanks for the improvement.

perobertson:
looks like a couple fixes didnt get made before my battery died. I can make those changes later tonight when im back online.

lib/cequel/metal/keyspace.rb
```diff
@@ -226,6 +227,32 @@ def execute_with_options(statement, options={})
...
+          client.prepare(cql)
+        rescue Cassandra::Errors::NoHostsAvailable,
+               Cassandra::Errors::ExecutionError
+          clear_active_connections!
+          raise if retries == 0
+          retries -= 1
+          sleep(retry_delay)
+          retry
+        end
+      end
```
pezra: Could we extract this and the execute retry logic into a retry method that takes a block?

perobertson:
dried up the retry code and fixed up the tests with the correct exceptions

pezra:
Thanks for the improvement!

perobertson:
Anytime! The next thing Ill probably look into is making this gem thread safe as it would be nice to use sidekiq.

pezra:
Thread safety would be awesome.

pezra:
Released in 2.0.2

PR #331, opened 2016-10-12
Fix new relic instrumentation

hydrogen18:
It looks like no one gave the new relic instrumentation any love in the 2.0.0 upgrade.

It fails for me in production, because the actual parameters are different in 2.0.0.

This fixes it for me.

### DIFF
--- a/lib/cequel/metal/new_relic_instrumentation.rb
+++ b/lib/cequel/metal/new_relic_instrumentation.rb
@@ -13,15 +13,28 @@ module Cequel
     module NewRelicInstrumentation
       extend ActiveSupport::Concern
 
-      define_method :execute_with_options_with_newrelic do |statement, bind_vars, options|
-        callback = Proc.new do |result, scoped_metric, elapsed|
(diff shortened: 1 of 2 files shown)


pezra:
Is there a test we could write so that does get missed next time?

hydrogen18:
Since this is basically a monkey patch we would have to do something bizarre inside a test, such as:

0. Stub out NewRelic
0. Save the old method as a reference
0. Require the file
0. Verify the function was monkey patch
0. Verify the arity of the monkey patch matches the new one
0. Verify the monkey patch calls the original method
0. Restore the original method

hydrogen18:
Let me know if that strategy sounds acceptable for a test, or outline another one. The test will likely be more brittle than the code it is covering.

pezra:
@hydrogen18, have a test that convoluted is unappealing (🙂 as, i suspect, was the intent of your description). Let's skip it for now.

pezra:
Released in 2.0.2

PR #335, opened 2016-10-20
expose if the dataset is on the last page

perobertson:
When working with paging you need to be able to determine if you are on the last page. If you are manually looping over a collection instead of using `.each` the `last_page?` comes in handy because `nil` is a valid paging state, it just means start at the beginning.

By exposing the paging state we are now able to write code like this:
```ruby
paging_state = nil
loop do
  posts = blog.posts.page_size(100).paging_state(paging_state)
  paging_state = posts.next_paging_state
  posts.approve!
  break if posts.last_page?
end
```

### DIFF
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -579,8 +579,24 @@ module Cequel
         end
       end
 
+      #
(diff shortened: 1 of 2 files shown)


pezra:

PR #336, opened 2016-11-01
Delegate error handling to a policy object, allow for arbitrary options to be passed to cassandra driver gem

hydrogen18:
After refactoring some of the keyspace object, I can't really figure out why you would want to call `clear_active_connections!` after each error. This may have worked around some shortcomings in the cassandra driver gem historically, but I don't think it is a very good idea. The cassandra driver gem supports an infinite amount of configuration in the form of policy objects, which can be used to implement whatever handling logic you would like. It also tracks what hosts are up and down, so there is little reason to clear the connections on an error. I also can't see why you would to do retries in this gem, given that the cassandra driver gem can do that for you.

What I've seen in production systems is that once our application encounters a single timeout, all application threads immediately start hammering all the nodes. I think this is because clearing the connections makes the driver gem "forget" which hosts are up and down. 

All that taken into consideration, I don't want to change the default behavior because that would make the version '3.0.0' basically. So instead I have added two options to the configuration hash passed to this gem

1. cassandra_error_policy
2. cassandra_options

Using these two options, you can disable the default behavior and pass whatever options you would like to the underlying driver gem. Specifically, specifying a behavior of `Cequel::Metal::Keyspace::Policy::CassandraError::Raise` results in errors from the cassandra driver gem simply propagating up the call stack.

Let me know if you need anything else to accept this merge.

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -413,10 +413,10 @@ module Cequel
         value = configuration.fetch(:cassandra_error_policy, ::Cequel::Metal::Policy::CassandraError::ClearAndRetryPolicy)
         # Accept a class name as a string, create an instance of it 
         if value.is_a?(String)
-          value.constantize.new
+          value.constantize.new(configuration)
         # Accept a class, instantiate it
         elsif value.is_a?(Class)
-          value.new 
+          value.new(configuration)
         # Accept a value, assume it is a ready to use policy object
         else 
           value
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -126,6 +126,10 @@ module Cequel
       #   key
       # @option configuration [String] :passphrase the passphrase for client
       #   private key
+      # @option configuration [String] :cassandra_error_policy A mixin for 
+      #   handling errors from Cassandra
+      # @option configuration [Hash] :cassandra_options A hash of arbitrary
+      #   options to pass to Cassandra
       # @return [void]
       #
       def configure(configuration = {})
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -36,6 +36,8 @@ module Cequel
       attr_reader :client_compression
       # @return [Hash] A hash of additional options passed to Cassandra, if any
       attr_reader :cassandra_options
+      # @return [Object] The error policy object in use by this keyspace 
+      attr_reader :error_policy
 
       #
       # @!method write(statement, *bind_vars)
@@ -139,7 +141,7 @@ module Cequel
         end
         @configuration = configuration
         
-        extend(extract_cassandra_error_policy(configuration))
+        @error_policy = extract_cassandra_error_policy(configuration)
         @cassandra_options = extract_cassandra_options(configuration)
         @hosts, @port = extract_hosts_and_port(configuration)
         @credentials  = extract_credentials(configuration)
@@ -324,8 +326,10 @@ module Cequel
           yield
         rescue Cassandra::Errors::NoHostsAvailable,
                Cassandra::Errors::ExecutionError,
-               Cassandra::Errors::TimeoutError => exc               
-          handle_error(exc, retries_remaining)
+               Cassandra::Errors::TimeoutError => exc
+          # Pass the error to the error policy. It can reraise,
+          # if it returns at all it indicates a retry should occur               
+          error_policy.handle_error(self, exc, retries_remaining)
           retries_remaining -= 1
           retry                    
         end
@@ -387,7 +391,11 @@ module Cequel
       end
 
       def extract_retry_delay(configuration)
-        configuration.fetch(:retry_delay, 0.5)
+        v = configuration.fetch(:retry_delay, 0.5)
+        if v <= 0.0
+          raise ArgumentError, "The value for retry must be a positive number, not '#{v}'"
+        end 
+        v
       end
 
       def extract_ssl_config(configuration)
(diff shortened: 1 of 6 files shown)


hydrogen18:
For reference, here is the full list of options you can pass to the cassandra driver gem

http://docs.datastax.com/en/developer/ruby-driver/3.0/api/cassandra/#cluster-class_method

pezra:
Fwiw, I have no problem with a version 3.0.0. Version are free, we can have as many as we need. That being said I do like not breaking the API unless it brings real benefits.

pezra:
I like the idea of moving the error handling into policy objects. How do you feel about having them be real objects, rather than mixins? We could have an abstract class for the interface and a `ResetAndRetryPolicy` to implement the current behavior.

hydrogen18:
They would need to have access to all the methods and instance variables of the `Keyspace` class, which is why I did not make them their own class.

pezra:
Why all the instance variables? I agree that keyspace would need to expose enough behavior to support error handling but that seems like a Good Thing to me.

hydrogen18:
The code does things like `sleep(retry_delay)`

pezra:
`Keyspace` already has too much responsibility. I am not comfortable adding the error handling policy, even via inheritance. A separate configuration object is one way we can handle this but for the short term i'm ok with `Keyspace` gaining simple accessors for configuration information. Once we have those methods as part of the public interface we can extract them into an object of their very own.

hydrogen18:
So the delgation will wind up looking something like ?

```
begin
  # stuff that can explode
rescue CassandraError => e
  error_policy.handle(self, e, retries_remaining)
  retry
```

The instance variable `error_policy` on `Keyspace` is always going to be defined, but will be a sane default if the user doesn't configure it.

hydrogen18:
I updated it with your requested refactor.

As a plus, some of the tests are a bit less weird.

lib/cequel/metal/keyspace.rb
```diff
@@ -311,17 +321,17 @@ def cluster
...
-               Cassandra::Errors::TimeoutError
-          clear_active_connections!
-          raise if retries == 0
-          retries -= 1
-          sleep(retry_delay)
-          retry
+               Cassandra::Errors::TimeoutError => exc
+          # Pass the error to the error policy. It can reraise,
+          # if it returns at all it indicates a retry should occur               
+          error_policy.handle_error(self, exc, retries_remaining)
```
pezra: It seems like retrying should be a feature of the error policy. Maybe sometimes one wouldn't want to retry. What about moving all of this functionality into the error policy as a `#execute_stmt(&blk)` method that would look very similar to the current `#client_retry`

lib/cequel/metal/keyspace.rb
```diff
@@ -393,7 +408,24 @@ def extract_ssl_config(configuration)
         ssl_config.each { |key, value| ssl_config.delete(key) unless value }
         ssl_config
       end
-
+      
+      def extract_cassandra_error_policy(configuration)
+        value = configuration.fetch(:cassandra_error_policy, ::Cequel::Metal::Policy::CassandraError::ClearAndRetryPolicy)
+        # Accept a class name as a string, create an instance of it 
+        if value.is_a?(String)
+          value.constantize.new
```
pezra: Maybe pass in `self` so the policy has access to the configuration information?

hydrogen18: Do you mean pass in `configuration` ?

pezra: Hmm... i don't love the idea of passing around a raw hash but maybe that is better than passing the keyspace?

hydrogen18: Yes, I feel that passing a reference to the 'parent' object is an  anti pattern.

Also, this happens during configuration. There is no guarantee that the keyspace is properly configured yet.

Maybe we could just pass a key of the `configuration` hash if it exists?

pezra: Let's pass `configuration` under the theory that real soon now we will improve it to be more than just a simple hash.

hydrogen18: Done.

hydrogen18:
Refactor is done.

lib/cequel/metal/policy/cassandra_error.rb
```diff
@@ -0,0 +1,79 @@
...
+              raise error if retries_remaining == 0
+              sleep(retry_delay)
+              keyspace.clear_active_connections! if clear_before_retry
+              retries_remaining -= 1
+              retry                    
+            end
+          end 
+        end
+        
+        module RetryPolicy
```
pezra: Should this be a class and inherit from `ErrorPolicyBase`?

lib/cequel/metal/policy/cassandra_error.rb
```diff
@@ -0,0 +1,79 @@
...
+          # See Keyspace#configure
+          
+          # On instantiation, the configuraiton hash passed to Cequel is 
+          # available here
+          def initialize(options = {})
+          end
+                    
+          def handle_error(error, retries_remaining)
+            raise RuntimeError, 'This is an abstract base class, never call this'
+          end
```
pezra: I think the above method should be replaced with
```ruby
def execute_stmt(keyspace)
  raise NotImplementedError, "#execute_stmt must be implemented in #{self.class}"
end
```

pezra:
Looks good except for the minor issues in my review. Thanks for the hard work.

lib/cequel/metal/policy/cassandra_error.rb
hydrogen18: If you look, I implemented `self.new` on this module. It actually returns an instance of `ClearAndRetryPolicy` that is configured to have `:clear_before_retry` set to false. This avoids code duplication.

This is a very Ruby-specific idiom, I'm not aware of any other languages that let you implement constructors with an arbitrary return values. Let me know if you think this needs to be changed.

hydrogen18:
I believe I have now addressed all your comments.

pezra:
Yay! Thanks for finishing that off @hydrogen18. I really appreciate the work you did on this PR.

pezra:

PR #337, opened 2016-11-07
skip synchronizing materialized view

emn178:
If I have a materialized view by creating manually and define the model, it will cause error when synchronizing. I want to skip it.
It's better to support define materialized view by cequel, but I think we can skip and prevent error for now.

### DIFF
--- a/lib/cequel/schema/table_reader.rb
+++ b/lib/cequel/schema/table_reader.rb
@@ -30,6 +30,16 @@ module Cequel
       end
 
       #
+      # Check if it is materialized view
+      #
+      # @param (see #initialize)
+      # @return (see #materialized_view?)
+      #
+      def self.materialized_view?(keyspace, table_name)
+        new(keyspace, table_name).materialized_view?
+      end
+
+      #
       # @param keyspace [Metal::Keyspace] keyspace to read the table from
       # @param table_name [Symbol] name of the table to read
       # @private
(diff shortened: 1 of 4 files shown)


lib/cequel/schema/table_reader.rb
```diff
@@ -150,18 +173,21 @@ def read_table_settings
...
+          .keyspace(keyspace.name)
+          .table(table_name.to_s)
+      end
 
-            fail(NoSuchKeyspaceError, "No such keyspace #{keyspace.name}") if
-              !cluster.has_keyspace?(keyspace.name)
+      def cluster
+        @cluster ||= begin
+          @cluster = keyspace.cluster
+          @cluster.refresh_schema
```
emn178: In original table_data do this, so I also do this. Any suggestion?

pezra: I don't think this is a show stopper. It does only happen the first time, due to the memoization.

emn178:
I tried to cached which keyspace refreshed, but it won't work if schema changed. This can not pass unit tests. I will try other methods later.

pezra:
@emn178, i hope you find time to get the tests passing. This is a good change, over all, and i'd like to include it.

pezra:
@emn178, this looks good. If you rebase it i'll merge it.

pezra:
merge as #346. Thanks for the improvement.,

pezra:

PR #340, opened 2016-12-06
Fixes README.md

jdiago:
The `Counting` section was inside of the `Upserts` section. This just moves `Counting` under `Upserts`.


pezra:
Thanks for the improvement!

PR #348, opened 2017-01-09
test as default rake task

phoet:
this is a small change that would make the repository follow the rule of least surprise for contributors (like myself).

when i'm forking a project and trying to provide a PR i usually run `bundle exec rake` as almost all libraries then run the test-suite as their default task.

thanks for the great library!


pezra:
👍 thanks for the improvement

PR #349, opened 2017-01-09
add IF EXISTS

phoet:
adding support for ` IF EXISTS ` conditionals in `DROP TABLE` and `DROP MATERIALIZED VIEW` as i just would have needed that.

would be great if you would consider adding this. if you would recommend any changes, please go ahead.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -3,8 +3,7 @@ source 'https://rubygems.org'
 gemspec
 
 group :debug do
-  gem 'debugger', '~> 1.6', :platforms => :mri_19
-  gem 'byebug', '~> 2.7', :platforms => [:mri_20, :mri_21]
+  gem 'byebug', '~> 2.7'
   gem 'pry', '~> 0.9'
 end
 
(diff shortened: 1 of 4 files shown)


pezra:
I'd love to have this but, as you may have noticed from the tests, we support c* 2.x. I'm open to options for how this might be made to support both new and old versions. However, I am not ready to give up support for the older versions just yet. (I have projects still on them. :) )

phoet:
@pezra would it be feasible to configure the tests to not be run with cassandra 2x and update the documentation to reflect support with 3x?

pezra:
I am not ready to drop support for the older versions just yet.

We could potentially detect which version of C* we are connected to and then only add the `IF EXISTS` when the C* version is high enough.

phoet:
oh sorry, i did not mean to drop tests for old versions in general, just for this functionality! usually the way i would go about this is to add meta-data to rspec and than have a tag on the spec so that it will only be run for a match (in this case the cassandra version).

phoet:
same goes for the documentation, i was just referring to the usage of that specific flag.

pezra:
OIC... A CQL version tag on specs for newer CQL features is a good idea!.

phoet:
@pezra added tagging for cassandra version, please take a look

pezra:
👍 i like that approach to excluding those tests. I'll merge once travis completes and then we can the tagging on #353.

pezra:
released in 2.1.0

Issue #208, opened 2015-01-12
Can't filter by more than one indexed column


yeudit:
I'm getting an 'Cequel::Record::IllegalQuery: Can't scope by more than one indexed column in the same query' exception.
I have a record class:

    class Space
      include Cequel::Record
      timestamps

      key :id, :uuid, :auto => true
      column :name,         :text,     :index => true
      column :type,          :text,      :index => true
    end

I'm trying to get all spaces with name => x and type => y. Example:
Space.where({name: 'name', type: 'type'})

This raises 'Cequel::Record::IllegalQuery: Can't scope by more than one indexed column in the same query' error.
I want to know how can I filter records by 2 or more columns?

Micka33:
# From the [cassandra documentation](http://www.datastax.com/documentation/cql/3.0/cql/cql_reference/select_r.html)
``` sql
CREATE TABLE ruling_stewards (
  steward_name text,
  king text,
  reign_start int,
  event text,
  PRIMARY KEY (steward_name, king, reign_start)
);
```
This query constructs a filter that selects data about stewards whose reign started by 2450 and ended before 2500. If king were not a component of the primary key, you would need to create an index on king to use this query:
```sql
SELECT * FROM ruling_stewards
  WHERE king = 'Brego'
  AND reign_start >= 2450
  AND reign_start < 2500 ALLOW FILTERING;
```
The output is:

```console
steward_name | king  | reign_start | event
--------------+-------+-------------+--------------------
      Boromir | Brego |        2477 |   Attacks continue
       Cirion | Brego |        2489 | Defeat of Balchoth

```

To allow Cassandra to select a contiguous ordering of rows, you need to include the king component of the primary key in the filter using an equality condition. The [ALLOW FILTERING](http://www.datastax.com/documentation/cql/3.0/cql/cql_reference/select_r.html#reference_ds_d35_v2q_xj__selAllowFilt) clause is also required.

## ALLOW FILTERING clause

When you attempt a potentially expensive query, such as searching a range of rows, this prompt appears:
```quote
Bad Request: Cannot execute this query as it might involve data
filtering and thus may have unpredictable performance. If you want
to execute this query despite the performance unpredictability,
use ALLOW FILTERING.
```
To run the query, use the ALLOW FILTERING clause. Imposing a limit using the LIMIT n clause is recommended to reduce memory used.

-----------

I guess the solution is to add `ALLOW FILTERING` to the request sent by cequel. however I can't find in the documentation how to do that.

[...]

PR #353, opened 2017-01-21
hack in allow_filtering

phoet:
for reference: https://github.com/cequel/cequel/issues/208#issuecomment-273941857

this is a working prototype for adding `ALLOW FILTERING` support. the way that i squeezed it into the existing codebase will make it work like `Post.allow_filtering!.where(title: 'Cequel 0')`. 

`allow_filtering!` will have to be called (like ie `consitency(:all)`) before calling `where`. this is mainly due to the `ArgumentError` that will otherwise be thrown in `filter_secondary_index`.

that does not feel very sexy to me though...

i'd like to get some early feedback before proceeding.

### DIFF
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -50,6 +50,7 @@ module Cequel
       attr_reader :query_consistency
       attr_reader :query_page_size
       attr_reader :query_paging_state
+      attr_reader :allow_filtering
 
       def_delegator :keyspace, :write_with_options
 
@@ -573,6 +574,12 @@ module Cequel
         end
       end
 
+      def allow_filtering!
+        clone.tap do |data_set|
+          data_set.allow_filtering = true
+        end
+      end
+
       def paging_state(paging_state)
         clone.tap do |data_set|
           data_set.query_paging_state = paging_state
@@ -645,6 +652,7 @@ module Cequel
           .append(*row_specifications_cql)
           .append(sort_order_cql)
           .append(limit_cql)
+          .append(allow_filtering_cql)
       end
 
       #
@@ -675,9 +683,17 @@ module Cequel
         end
       end
 
+      # @private
+      def allow_filtering_cql
+        if allow_filtering
+          ' ALLOW FILTERING'
+        else ''
+        end
+      end
+
       protected
 
-      attr_writer :row_limit, :query_consistency, :query_page_size, :query_paging_state
+      attr_writer :row_limit, :query_consistency, :query_page_size, :query_paging_state, :allow_filtering
 
       private
 
--- a/lib/cequel/metal/data_set.rb
+++ b/lib/cequel/metal/data_set.rb
@@ -574,6 +574,9 @@ module Cequel
         end
       end
 
+      #
+      # @see RecordSet#allow_filtering!
+      #
       def allow_filtering!
         clone.tap do |data_set|
           data_set.allow_filtering = true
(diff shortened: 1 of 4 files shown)


phoet:
i was wondering if it might be possible to implement this via a global configuration setting ie `Cequel.allow_filtering!`.

with this, i could enable it for ie. a debugging session or in the development environment or if "i know what i'm doing".

i understand your concerns for trying to protect people from shooting themselves in the foot, but i don't see any difference to doing `Redis.keys` or a sql-query with no index...

lib/cequel/metal/data_set.rb
```diff
@@ -573,6 +574,12 @@ def page_size(page_size)
         end
       end
 
+      def allow_filtering!
```
pezra: I think a method comment would be helpful to explain what this is for and which C* version support it.

lib/cequel/record/data_set_builder.rb
```diff
@@ -99,6 +101,12 @@ def set_consistency
         end
       end
 
+      def set_allow_filtering
```
pezra: method docs here too

lib/cequel/record/record_set.rb
```diff
@@ -465,6 +465,16 @@ def consistency(consistency)
       end
 
       #
+      # Set the consistency at which to read records into the record set.
+      #
+      # @param consistency [Symbol] consistency for reads
```
pezra: This line is incorrect.

lib/cequel/record/record_set.rb
```diff
@@ -465,6 +465,16 @@ def consistency(consistency)
...
+      # @return [RecordSet] record set tuned to given consistency
+      #
```
pezra: Let's add a compatibility note here, too.

spec/examples/record/record_set_spec.rb
```diff
@@ -805,6 +805,12 @@ def self.latest(count)
           to raise_error(ArgumentError)
       end
     end
+
+    context 'allow_filtering!' do
```
pezra: Tag this example so it will be skipped on old versions.

phoet:
@pezra sorry, dont want to sound rude, did you actually read my comments?

pezra:
@phoet, wondering if i read your comment is a fair question since i didn't respond directly.

Enabling `allow filtering` globally is much more dangerous than enabling it just for individual queries. In my experience `allow filtering` doesn't work very well for non-trivially sized tables. I am worried that those new to C* will enable allow filtering globally because it allows C* to behave more like a relational database. Doing so would probably be fine in development but will fail miserably if they take the app to production. 

I see the appeal for debugging and troubleshooting but i think the risk is too high for the limited benefit. Particularly, when the same effect is accomplished by this PR at much lower risk.

pezra:
if you add the documentation to the methods and readme and get the build green i'd merge this today so that it is in the release i am planning this afternoon.

lib/cequel/record/data_set_builder.rb
phoet: all those methods are without doc, the builder is kind of a private api

phoet:
@pezra i added some documentation. unfortunately i have to run now, so can't see if the build turns green or any other feedback. feel free to adapt anything that is not to your liking.

lib/cequel/record/data_set_builder.rb
pezra: i don't think internality is a particularly compelling excuse for omitting method docs. However, it is a small thing.

pezra:
Merged. Thanks.

pezra:
release in 2.1.0

phoet:
<3

PR #355, opened 2017-01-23
fix documentation to have 2 args

phoet:
was copying the method signature from the docs, but it does only take 2 args now.

i'm unsure about updating the rest of the docs, looks like quite a lot has changed there...


pezra:
👍  thanks for the improvement

PR #359, opened 2017-01-27
Fix ignoring options

Eraden:
If keyspace configuration is ActiveSupport::HashWithIndifferentAccess options will ignore keys which exists.


pezra:
Thanks for the improvement.

Issue #360, opened 2017-02-03
::Cassandra::Statements::Batch is breaking NewRelic


lxbrito:
Cequel::Metal::NewRelicInstrumentation is passing statement_txt as nil when the statement is a  ::Cassandra::Statements::Batch and thus breaking the NewRelic Gem when it tries to truncate statement_txt.

shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/transaction_sampler.rb:155:in `truncate_message': undefined method `length' for nil:NilClass (NoMethodError)
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/transaction_sampler.rb:143:in `notice_extra_data'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/transaction_sampler.rb:220:in `notice_nosql_statement'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/datastores.rb:185:in `notice_statement'
    from shared/bundle/ruby/2.3.0/gems/cequel-2.1.0/lib/cequel/metal/new_relic_instrumentation.rb:31:in `block (2 levels) in <module:NewRelicInstrumentation>'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/datastores.rb:115:in `block in wrap'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/method_tracer.rb:73:in `block in trace_execution_scoped'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/method_tracer_helpers.rb:82:in `trace_execution_scoped'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/method_tracer.rb:71:in `trace_execution_scoped'
    from shared/bundle/ruby/2.3.0/gems/newrelic_rpm-3.15.2.317/lib/new_relic/agent/datastores.rb:108:in `wrap'
    from shared/bundle/ruby/2.3.0/gems/cequel-2.1.0/lib/cequel/metal/new_relic_instrumentation.rb:48:in `block in <module:NewRelicInstrumentation>'
[...]

PR #361, opened 2017-02-20
#360

lxbrito:
Solves Issue #360 
::Cassandra::Statements::Batch is breaking NewRelic


pezra:
Thanks!

pezra:
will be in v2.1.1

PR #364, opened 2017-02-21
don't set table name when it is already present

### LINKED ISSUE #363: cequel:migrate not making tables with custom names
even when i have table_name defined in model `self.table_name = :custom_name` cequel migrate creates the table with name derived from the model

RajatGoyal:
this PR fixes #363


pezra:
will be in v2.1.1

PR #365, opened 2017-03-07
Code Climate config

larkinscott:
Add Code Climate config with Rubocop and duplication enabled.


pezra:
@larkinscott, thanks for the PR.

PR #378, opened 2017-09-13
Fix noisy logs in Ruby 2.5, ActiveModel 5.2 support

wflanagan:
This pull request fixes a duplicate column names issue we are seeing in using Cequel in ActiveJob/Sidekiq.  

Over 100k jobs, we see about 300 jobs where after the Sidekiq fork, the code errors with failure in the "finder methods"... where it has built a "find_by_id_and_id" method.  These jobs never correct, and continue to fail over and over again.  This is the same error we saw with our use of rspec-preloader in the 1.x.x branch of Cequel.  

This patch fixes it by turning the storage of column names from simple arrays into hashes.  Then, it uses the last value supplied to this as the value for production. This eliminates the ability for the column names to be duplicated on a reload.  

In addition, this pulls in the "Hornet Network" (thanks @pascalh1011) fix for noisy logs with Ruby 2.4.  I based from theirs and implemented my fixes. 

I've also incorporated "BMurray's Pull Request" for ActiveModel 5.2 support in (thanks @bmurray) to provide ActiveModel 5.2 support. 

And, lastly, I've updated a few cases of the def_delegator(s) to implement a method_missing approach to handle these cases, so we can bypass the noise of the deprecated Forwarable message. Note that in doing this, it bypasses the whole "private" thing all together by implementing the calls using method_missing.


wflanagan:
FYI.. this tests green on my local.

wflanagan:
FYI.. hold on this for a bit.. we found another case where this still has an error.

pezra:
@wflanagan,  thanks for improvements. i looked at the travis build and it seems that the tests are failing against cassandra 2.2 and 3.0. they pass against 2.1. the version of ruby doesn't seem to matter.

wflanagan:
Strange. I'm running Cassandra 3.0 locally.

wflanagan:
@pezra I looked at the Travis and the errors are related to lists. Given that I was trying to fix the other bug that the person mentioned with lists, I looked at that (i.e. touched that code putting puts in) but THOUGHT I took it all out and reset to a similar state.  

I'll go through the Travis logs and see if I can fix those errors.

wflanagan:
FYI.. i'll get back to this soon.. had to take a detour through other parts of the app for a bit..

wflanagan:
@pezra FYI.. this is green now.

pezra:
@wflanagan, 👍. Can you rebase?

wflanagan:
Ok.. That's my first time "rebasing" through a fork.  So, I'm not 100% sure I did that correctly.. but its on its way up... :-)

lib/cequel/schema/table.rb
```diff
@@ -200,11 +256,10 @@ def compact_storage?
 
       protected
 
-      attr_reader :columns_by_name
+      # attr_reader :columns_by_name
```
pezra: Remove if unneeded.

spec/examples/record/callbacks_spec.rb
```diff
@@ -35,12 +35,12 @@ def executed_callbacks
...
     attr_writer :instance_after_save
 
-    private
-
     def create_post
       post = Post.create!(permalink: 'autopost', title: 'Auto Post')
       self.post = post
     end
+    
+    private
```
pezra: Does `#create_post` need to be public?

erkie:
@wflanagan @pezra this issue is really killing our logs. Can I jump in and fix these comments somehow?

pezra:
@erkie, please do

erkie:
Had a look at the code, turns out it's way above my head :(

wflanagan:
FYI, I've updated my original pull request to handle Ruby 2.5/2.4 noisy logs, ActiveModel 5.2, and my original fix to avoid duplicated finder methods.  I've also added a Dockerfile and docker-compose.yml so that you can do development without having to setup anything specific with Cassandra.

wflanagan:
FYI.. has the .has_rdoc fix in it as well.

wflanagan:
Checking on this.. don't want it to go bad.. :-)

erkie:
@pezra maybe a mention will help? :)

alxgsv:
I use this pull request in production for 1 month now. Thank you, @wflanagan!

wflanagan:
@pezra Thought I'd check/ping on this again.

pezra:
@wflanagan can you address the other comments from my review?

wflanagan:
Hey Peter,

I think I addressed most of them.  That review was more than a year ago..
and I made many of those fixes before.  At the time, it wasn't passing
certain versions of Ruby etc.  Now, it is.

Is there something else you want added/modified?

Wf

ilanusse:
Hello, is there anything I can do to help this roll along? I'm working on a Rails 5.2 project and would love to use Cequel for it.

wflanagan:
So, I think Peter wants me to make changes to the codebase. But it's been
MORE than a year since I submitted them. Some of the changes he requested,
I think, were fixed. But, some are still outstanding.

IF I can get a punch list of the requested changes, what's below the line
and HAS to be fixed, I guess I can try to push this out. But, a lot of what
is being asked or shown as outstanding really is in the "I have no idea"
category at this point, and I just have to run the tests and see.

And, given that we've been running this in production for quite some time,
as have others, I'm franking at this point a bit hesitant to break it by
making a bunch of items private, etc.

But, if no one else can step in and confirm these changes, I guess i'll try
to make it a priority.

Let me know.

pezra:
@wflanagan outstanding issues for this PR:

- [] https://github.com/cequel/cequel/pull/378#discussion_r155902760
- [] https://github.com/cequel/cequel/pull/378#discussion_r155902874
- [] https://github.com/cequel/cequel/pull/378#discussion_r155903209
- [] https://github.com/cequel/cequel/pull/378#discussion_r155914135

If we reach concensus on those i'll merge this.

erkie:
@ilanusse this PR has been open for so long that any new energy is probably appreciated. Please do!

lib/cequel/schema/table.rb
```diff
@@ -52,24 +51,81 @@ def materialized_view?
...
-                        when ClusteringColumn
-                          @clustering_columns
-                        else
-                          @data_columns
-                        end
+        unless columns_by_name[column_desc.name]
+          column_description_store(column_desc)[column_desc.name] = column_desc
+          columns_by_name[column_desc.name]                       = column_desc
+        end
+      end
```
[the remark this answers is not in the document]
ilanusse: I believe Cassandra overwrites in this case, so let's go with last in.

lib/cequel/record/persistence.rb
```diff
@@ -260,6 +262,16 @@ def hydrate(row)
...
+      def updater
+        raise ArgumentError, "Can't get updater for new record" if new_record?
+        @updater ||= Metal::Updater.new(metal_scope)
+      end
+
+      def deleter
+        raise ArgumentError, "Can't get deleter for new record" if new_record?
+        @deleter ||= Metal::Deleter.new(metal_scope)
+      end
+
```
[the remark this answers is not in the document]
ilanusse: screen shot 2018-10-19 at 15 59 59

Apparently, these two methods cause lots of warnings because of all the delegations being made unless they're public.

wflanagan:
That sounds right.. that's why I moved them to public.

lib/cequel/record/persistence.rb
```diff
@@ -82,7 +82,7 @@ def hydrate(row)
       # @since 1.0.0
       #
       def key_attributes
-        @attributes.slice(*self.class.key_column_names)
+        @cequel_attributes.slice(*self.class.key_column_names)
```
pezra: why the name change?

lib/cequel/record/persistence.rb
```diff
@@ -185,8 +185,10 @@ def loaded?(column = nil)
       #
       def save(options = {})
         options.assert_valid_keys(:consistency, :ttl, :timestamp)
-        if new_record? then create(options)
-        else update(options)
+        if new_record?
+          create(options)
+        else
+          update(options)
```
pezra: 👍

lib/cequel/schema/table.rb
```diff
@@ -13,20 +13,20 @@ class Table
...
 
       # @return [Array<ClusteringColumn>] clustering columns defined on the
       #   table
-      attr_reader :clustering_columns
+      # attr_reader :clustering_columns
 
       # @return [Array<DataColumn,CollectionColumn>] data columns and
       #   collection columns defined on the table
-      attr_reader :data_columns
+      # attr_reader :data_columns
```
pezra: let's delete these if they aren't needed anymore.

lib/cequel/record/persistence.rb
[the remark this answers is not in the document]
wflanagan: That's correct and why I did it. Otherwise, it was a MUCH larger refactor.

lib/cequel/schema/table.rb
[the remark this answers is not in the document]
wflanagan: @pezra Does that suffice or do you want to do something else?

ilanusse:
Created a new PR (#397) because @wflanagan mentioned having production sites running on his fork code, so I didn't want to mess around with his fork.

lib/cequel/schema/table.rb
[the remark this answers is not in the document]
pezra: I'd prefer to see this method fail on repeated columns. Repeats seem likely to be bugs, rather than intended behavior.

lib/cequel/schema/table.rb
```diff
@@ -52,24 +51,81 @@ def materialized_view?
...
-        columns_by_name[column_desc.name] = column_desc
+      # Stores the column description in the appropriate storage hash
+      # for the column name and column type.
+      def column_description_store(column_desc)
+        value = {
+          Cequel::Schema::PartitionKey     => partition_key_columns_hash,
+          Cequel::Schema::ClusteringColumn => clustering_columns_hash
+        }.try(:[], column_desc.class)
+        value = data_columns_hash if value.nil?
+        value
```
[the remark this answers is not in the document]
wflanagan: changed in my local branch.. will push in a bit.

lib/cequel/record/data_set_builder.rb
```diff
@@ -122,6 +122,14 @@ def set_paging_state
...
         ascends_by?(order_by_column) ? :asc : :desc
       end
+
+      def method_missing(m, *args, &block)
+        if record_set_delegated_methods.include?(m)
+          record_set.send(m, *args, &block)
+        else
+          super
+        end
+      end
```
pezra: `#method_missing` is *much* slower than straight delegation. Why change?

lib/cequel/record/lazy_record_collection.rb
```diff
@@ -77,12 +74,21 @@ def assert_fully_specified!
...
-
       def key_attributes_for_each_row
         map { |record| record.key_attributes }
       end
+
+      def record_set_delegated_methods
+        %i[table connection key_columns scoped_key_values]
+      end
+
+      def method_missing(m, *args, &block)
```
pezra: `#method_missing` is much slower than straight delegation. Why change?

lib/cequel/record/persistence.rb
ilanusse: When `@attributes` is exposed in Rails 5.2 it causes problems

spec/examples/record/callbacks_spec.rb
wflanagan: I don't remember moving this code (maybe I did).. so I moved it back.

wflanagan:
FYI.. i'm pushing a version momentarily that fixes these comments.

lib/cequel/schema/table.rb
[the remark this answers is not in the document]
wflanagan: Is there an existing test of this case?  FYI, the old one, I don't think, did this (failing).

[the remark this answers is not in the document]
pezra: I doubt there is an existing test. There must not be since this branch has a green build.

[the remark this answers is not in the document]
wflanagan: Adding a test case, and making this return an ArgumentError

lib/cequel/record/persistence.rb
pezra: What problems does it cause? See <https://github.com/cequel/cequel/pull/395#discussion_r226765188> for detailed discussion.

wflanagan:
@pezra @ilanusse Just pushed fixes for all requested items. I also fixed the Fixnum deprecation warning on Ruby 2.5.

lib/cequel/schema/table.rb
wflanagan: ok.. i'll push this change.

lib/cequel/record/data_set_builder.rb
pezra: If the problem was the warning about delegating to a private method we could convert the target method to `protected`. Protected is generally a better choice than private anyway, IMO.

lib/cequel/record/lazy_record_collection.rb
wflanagan: The delegation is to a private method, and this throws off MASSIVE logs. This is a part of this change. If it's reimplemented as a delegation, then it will start to throw a LOT of messages about delegating to a private method.  

This approach keeps the messages from happening, and still raises on the non-defined methods. 

I pulled this approach from another gem that had the same problem... this approach was how they solved it.

lib/cequel/record/data_set_builder.rb
wflanagan: For the schema/table.rb note, it now fails on repeated columns. I added a test to test this case as well.

wflanagan: Would have been great to get this feedback about 8 months ago. :-) Alright.. i'll rework the entire approach to try to make it work.  Ugh. :-(

wflanagan:
Alright. I won't get to this today.. and not sure when.

wflanagan:
@pezra I've tested this, and it seems that it still throws the same warning logs using protected as it does with private...

lib/cequel/record/data_set_builder.rb
pezra: Sorry i didn't see this until today. I changes i did request didn't get done until today, though.

If we need access to those method from another class then they should be public. Period. Use `def_delegate` to bypass method access control is a terrible idea. Let's just make them public and call it a day.

lib/cequel/record/lazy_record_collection.rb
pezra: Let's just make the target methods public. See https://github.com/cequel/cequel/pull/378#discussion_r226803027

wflanagan:
@pezra Alright, I did it a different way that avoids using method_missing. Please take a look and see if that way works better for you.  

The change to protected DID NOT solve the problem.  The messages still print in Ruby 2.4+. So, instead, I created methods explicitly that were the ones being delegated.  

This might perform better. But, IMO, the code isn't as maintainable.. and I can think of some edge cases (related to blocks) where things might break. However, evidently, in Rails, this is how they were handling the issue for backward compatibility.  So, maybe that's a good solution for you? 

Or, would you like something done another way?

lib/cequel/record/data_set_builder.rb
wflanagan: I guess this boils down to a difference in our respective development cultures. 

I've already, based on your last comment, reimplemented them in a way that doesn't call the method_missing at all or use delegation, by exposing them as method directly on the classes.  I had already started this before I saw your comment above (stayed up late last night to implement this to get this finally done). 

If this is acceptable, then OK. If not, let me know.. as I'm not super excited about making another significant change to the code base. 

On your other comments about changes not being done until yesterday, a few thoughts for the future. 

With all the other changes made, and that **I did go through and remove comments**, but evidently missed a few, I wouldn't have expected that to hold up these changes for as long as it has.  I was under the impression that I had removed all the commented out code. If you were waiting on me to remove the 3 or 4 more lines of code, then it would have been great for you to tell me that, and point me to where you found them for me to remove. Or, even better, just remove them yourself. That "lag" of 8 months over 3 or 4 lines of commented out code seems like a waste.

And, on the other issue (add_column erroring versus taking the last one in), I didn't perceive your comment as an absolute request--more of a question about how it was implemented. We're talking about a method that had 3 or 4 lines of code. If there was a needed change, it wouldn't have been hard to fix.. and all of this could have been in production 8+ months ago. 
Anyway, it's been changed to raise an error now, and I added a test for it to ensure that it does. 

Anyway, it's different culture, and all good. 

Let me know about the delegation stuff.

pezra:
Closing in favor of #401

PR #379, opened 2017-11-01
added new data type :date

ballcheck:


pezra:
@ballcheck, thanks for the contribution. These failures don't look related to your change. I'll take a look an see if i can figure out what the deal is.

pezra:
#381 should fix the build

PR #380, opened 2017-11-08
Fix parsing of uuids from strings

mkonikowski:
Before this fix, range query argument value was compared with `Cassandra::Uuid`, so it was not possible to pass `timeuuid` as a string. It was not packed in mentioned class and cequel gem tried to parse it as a `Time`, which was giving `ArgumentError: argument out of range`. 

```
          elsif column.type?(:timeuuid) && !Cequel.uuid?(value)
            TimeuuidBound
```

https://github.com/cequel/cequel/blob/master/lib/cequel/record/bound.rb#L27-L38

Now queries that have `timeuuid` in string argument are working. Example from specs:

```
> before = Cassandra::Uuid::Generator.new.now.to_s
 => "31974816-c49e-11e7-b55e-1b9ba15056b7"
> PublishedPost['cassandra'].before(before)
=> # records published before  2017-11-08 16:02:17 UTC
```

### DIFF
--- a/lib/cequel/uuids.rb
+++ b/lib/cequel/uuids.rb
@@ -34,10 +34,18 @@ module Cequel
     # @return [Boolean] true if the object is recognized by Cequel as a UUID
     #
     def uuid?(object)
+      return true if uuid_in_string?(object)
+
       object.is_a?(Cassandra::Uuid)
     end
 
     private
+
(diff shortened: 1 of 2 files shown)


pezra:
#381 should fix the build. Once we get a green build for this i'll merge it. 

Thanks for the improvement.

PR #395, opened 2018-06-29
ActiveModel 5.2 Support

bmurray:
This is based on the existing pull request #392, and includes a fix for the Travis-CI build errors against ActiveModel 4.0. 

Alternatively, one could just drop support for 4.0.

### DIFF
--- a/lib/cequel/record/persistence.rb
+++ b/lib/cequel/record/persistence.rb
@@ -82,7 +82,7 @@ module Cequel
       # @since 1.0.0
       #
       def key_attributes
-        @attributes.slice(*self.class.key_column_names)
+        @cequel_attributes.slice(*self.class.key_column_names)
       end
 
       #
@@ -160,7 +160,7 @@ module Cequel
       # @since 1.0.0
       #
       def loaded?(column = nil)
-        !!@loaded && (column.nil? || @attributes.key?(column.to_sym))
+        !!@loaded && (column.nil? || @cequel_attributes.key?(column.to_sym))
       end
 
       #
@@ -361,17 +361,17 @@ module Cequel
       end
 
       def attributes_for_create
-        @attributes.each_with_object({}) do |(column, value), attributes|
+        @cequel_attributes.each_with_object({}) do |(column, value), attributes|
           attributes[column] = value unless value.nil?
         end
       end
 
       def attributes_for_update
-        @attributes_for_update ||= {}
+        @cequel_attributes_for_update ||= {}
       end
 
       def attributes_for_deletion
-        @attributes_for_deletion ||= []
+        @cequel_attributes_for_deletion ||= []
       end
 
       def assert_keys_present!
--- a/lib/cequel/record/dirty.rb
+++ b/lib/cequel/record/dirty.rb
@@ -47,7 +47,12 @@ module Cequel
       def save(options = {})
         super.tap do |success|
           if success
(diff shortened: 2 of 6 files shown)


wflanagan:
@bmurray I added your changes to my pull request, which also fixes noisy logs in Ruby 2.5.  FYI.

lib/cequel/record/persistence.rb
```diff
@@ -82,7 +82,7 @@ def hydrate(row)
       # @since 1.0.0
       #
       def key_attributes
-        @attributes.slice(*self.class.key_column_names)
+        @cequel_attributes.slice(*self.class.key_column_names)
```
pezra: Why the name change?

wflanagan: If i recall, there was a name conflict, where the "attributes" term was exposed, and it caused problems.

wflanagan: I wouldn't have changed it unless there was a reason...

ilanusse:
Also covered by #397.

lib/cequel/record/persistence.rb
pezra: @ilanusse, what is "pretty much the case"?

It seems from <https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/attributes.rb#L70>, <https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/dirty.rb#L137>, etc that active model still treats `@attributes` specially. I'm worried that this change with break some of the built in activemodel code.

I think this change came from  #395. @bmurray, can you shed light on this?

wflanagan: What would the test case look like that makes you comfortable with that?

pezra: Fair question. Not sure. I suppose the existing test suite should be enough.

I'd really like to understand the original motivation for the change, though.

wflanagan: At this point (it's been more than a year).. the only way to know for sure is to "undo" that change.. run the tests, and see what happens. I guess I could do that?

pezra: I've done it. I'll write it up in a min.

wflanagan: So, do I need to revert to that? Does it function, or does it have other problems? FYI.. my branch, with the other changes, as passed al the tests on Travis.. FYI.

pezra: So... [A year ago ActiveModel inherited better attribute handling from ActiveRecord.](https://github.com/rails/rails/commit/c3675f50d2e59b7fc173d7b332860c4b1a24a726#diff-8c369514b6ba036b727449c02e58e0fb) As of that change `@attributes` needs to be an [`AttributeSet`](https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/attribute_set.rb) of [`Attributes`s](https://github.com/rails/rails/blob/5-2-stable/activemodel/lib/active_model/attribute.rb) rather than a simple `Hash`. It looks like they maintained backwards compatibility with `attr_accessor` based active models by triggering the new behavior off of the existence of an `@attributes` instance variable.  

Changing our ivar name allows us to access the old behavior. It would be *way* better to adopt the new attribute approach than to simply change our instance variable name but i suppose that is a lot to ask.

PR #396, opened 2018-07-18
Remove deprecated has_rdoc

jdelStrother:
has_rdoc has been deprecated for 9 years now.  It causes a warning during `bundle install` - "Gem::Specification#has_rdoc= is deprecated with no replacement. It will be removed on or after 2018-12-01".

I think it was first deprecated here https://blog.rubygems.org/2009/05/04/1.3.3-released.html, and rubygems started warning about it here https://blog.rubygems.org/2011/03/28/1.7.0-released.html

How about removing it?


pezra:
Thanks!

PR #397, opened 2018-10-19
Support for Rails 5.2 and fixes noisy logs in Ruby 2.5

ilanusse:
## Summary
So I added some finishing touches on great work by @wflanagan, @bmurray, and @erkie in PR #378.
This should be enough to cover both that PR and #395.

## Features
- Removes noisy logs in Ruby 2.5 by changing a few protection levels. I played around with them to see  if changing protection levels could be avoided but no such luck. This seems the way to go.
- Adds support for Rails 5.2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,17 @@
+## 3.0.4
+* Add Dockerfile and docker-compose.yml for use in development of the gem.
+* Fix noisy logs from Ruby 2.5, ActiveModel 5.2 compatibility.
+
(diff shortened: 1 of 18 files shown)


ilanusse:
fwiw, this PR is pretty much the same as #378 and it seems the discussion over there has sparked more activity in that PR. We can leave this one alone for now

ilanusse:
@pezra You can close this if you'd like! Thanks for merging #395 so I can use Cequel for my project!

PR #401, opened 2018-10-23
fix calls that are producing many annoying warning log messages in modern ruby

pezra:
This is based on @wflanagan's work in #378. 

This still has on delegation to a private method. Make `RecordSet#connection` causes the test suite to hang and i don't have time to track down why. Even with this is much better than the current state of master.


pezra:
released as v 3.0.4

PR #404, opened 2018-11-28
Can now query with ALLOW FILTERING for multiple non-indexed columns

ilanusse:
## Summary
Can now query for more than one non-indexed column when allow filtering is active.

## Reason
Ran into a problem where I was looking to filter by more than one column within a table. Since most partitions would have only a few hundred tuples, I decided to use allow filtering, since it probably wouldn't lead to memory failure. I am backed by `cassandra-stress` results. 

I tried to query this way through Cequel but I got an `IllegalQuery` exception. I tried it out in CQL:

screen shot 2018-11-28 at 13 24 16
screen shot 2018-11-28 at 13 20 44

As you can see, once allow filtering is declared, then queries by more than one non-indexed columns are legal in CQL.

So I decided to add this in Cequel! Hence this PR.

### DIFF
--- a/lib/cequel/record/record_set.rb
+++ b/lib/cequel/record/record_set.rb
@@ -846,7 +846,7 @@ module Cequel
                "Can't scope key column #{column_name} without also scoping " \
                "#{missing_column_names.join(', ')}"
         end
-        if scoped_indexed_column
+        if scoped_indexed_column && !allow_filtering
           fail IllegalQuery,
                "Can't scope by more than one indexed column in the same query"
         end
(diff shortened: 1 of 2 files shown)


ilanusse:
Force-pushed after rebasing

ilanusse:
@pezra What do you think about this PR?

pezra:
I haven't had a chance to look at this yet. I'll try to get to it next week.

pezra:
@ilanusse thanks for the improvement!

pezra:
released as v3.2.0. 

Thanks!

PR #407, opened 2019-02-22
Fixed a bug where multiple non-indexed column filtering was not working

ilanusse:
## Summary
Remember https://github.com/cequel/cequel/pull/404/? It turns out it was badly tested and thus I couldn't see that the filters would replace the previous filter since they weren't merging together in `filter_secondary_index` (which I renamed to something I feel is more accurate: `filter_secondary_column`).

So I found the bug while doing QA on a project of mine, here's the fix.

I'm sorry I haven't been able to work on the UDT PR but things have been hectic at work these past few weeks. I'll have some time in a couple of weeks.

Cheers!

### DIFF
--- a/lib/cequel/record/record_set.rb
+++ b/lib/cequel/record/record_set.rb
@@ -107,7 +107,7 @@ module Cequel
 
       # @private
       def self.default_attributes
-        {scoped_key_values: [], select_columns: []}
+        { scoped_key_values: [], select_columns: [], scoped_secondary_columns: {} }
       end
 
       # @return [Class] the Record class that this collection yields instances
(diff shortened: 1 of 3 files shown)


ilanusse:
Force-pushed to rebuild! Seems to have worked 👌

lib/cequel/record/record_set.rb
```diff
@@ -846,15 +846,16 @@ def filter_secondary_index(column_name, value)
...
+               "Can't scope by more than one indexed column in the same query without allow_filtering!"
         end
         unless column.indexed? || allow_filtering
           fail ArgumentError,
-               "Can't scope by non-indexed column #{column_name}"
+               "Can't scope by non-indexed column #{column_name} without allow_filtering!"
         end
-        scoped(scoped_indexed_column: {column_name => column.cast(value)})
+        filter = { column_name => column.cast(value) }
+        scoped(scoped_secondary_columns: scoped_secondary_columns ? scoped_secondary_columns.merge(filter) : filter)
```
pezra: The logic in the method is really complicated. what if we reorganized it like

```
       if allow_filtering
         # bypass safety checks

       elsif scoped_secondary_columns.any?
          fail IllegalQuery,
               "Can't scope by more than one indexed column in the same query without allow_filtering!"

        elsif ! column.indexed?
          fail ArgumentError,
               "Can't scope by non-indexed column #{column_name} without allow_filtering!"
        end

        scoped(scoped_secondary_columns: 
               scoped_secondary_columns.merge(column_name => column.cast(value)) )
```

We would need to provide default value of `{}` for `scoped_secondary_columns` but i think that is better anyway. It allows the rest of the code, like this, to be less timid.

ilanusse: @pezra I am not sure how to set the default value for `scoped_secondary_columns`, where should I look?

ilanusse: Does sth like this look ok? I'm just missing the default value for `scoped_secondary_columns`

```ruby
def filter_secondary_column(column_name, value)
  column = target_class.reflect_on_column(column_name)
  validate_secondary_column_filter(column)
  scoped(scoped_secondary_columns:
    scoped_secondary_columns.merge(column_name => column.cast(value)))
end

def validate_secondary_column_filter(column)
  if column.nil?
    fail ArgumentError,
      "No column #{column_name} configured for #{target_class.name}"
  end
  if column.key?
    missing_column_names = unscoped_key_names.take_while do |key_name|
      key_name != column_name
    end
    fail IllegalQuery,
      "Can't scope key column #{column_name} without also scoping " \
      "#{missing_column_names.join(', ')}"
  else
    validate_non_key_column_filter(column)
  end
end

def validate_non_key_column_filter(column)
  return if allow_filtering

  if scoped_secondary_columns.any?
    fail IllegalQuery,
      "Can't scope by more than one indexed column in the same query without allow_filtering!"
  end
  unless column.indexed?
    fail ArgumentError,
      "Can't scope by non-indexed column #{column_name} without allow_filtering!"
  end
 end

ilanusse: @pezra Nevermind! Found it! Code updated!

ilanusse: @pezra Hello! Any chance you could look at the updated PR, please?

pezra:
released as 3.2.1. thanks for the improvements!

PR #412, opened 2019-05-20
Ruby 2.6 BigDecimal initialization compatibility

alxgsv:


lib/cequel/type.rb
```diff
@@ -282,7 +282,7 @@ def cast(value)
     #
     class Decimal < Base
       def cast(value)
-        value.is_a?(BigDecimal) ? value : BigDecimal.new(value, 0)
+        value.is_a?(BigDecimal) ? value : BigDecimal(value, 0)
```
pezra: Do we even need the ternary? Most of the `<class>()` methods handle being given a value of that target type gracefully.

alxgsv:
Yes, I've been in process of upgrading my app to Ruby 2.6.3 and Rails 6. Fixed code spams _a lot_ of warnings like:
```
warning: BigDecimal.new is deprecated; use BigDecimal() method instead
```
BigDecimal constructor is removed: https://github.com/ruby/bigdecimal/blob/master/CHANGES.md
`BigDecimal()` however works on all supported Ruby versions (I've tested 1.8.7)

So I decided to just replace it and it worked.

lib/cequel/type.rb
alxgsv: Honestly I don't know how `BigDecimal()` works inside, and probably shouldn't. Maybe it's idempotent. Maybe it doesn't do extra work in this case. Maybe it will work like this forever. Too much assumptions for me :-)
On the other hand I find current code straightforward and I understand decision making process behind it.

wflanagan:
@alxgsv Hey, where are you on this? I'm trying to do this process (Rails 6 upgrade) now. So far, I"ve not swung back to this version (still using my own version that forked for the code I wrote).  But, i'm just starting now. Is it working?  What needs to be fixed? If I get in and fix it, can be push a "Rails 6 compatible" version?

alxgsv:
@wflanagan I used this commit in production with Rails 6.0.0.rc1 and Rails 6.0.0.rc2 with no issues. This commit is required by Ruby 2.6.3 not Rails 6. So, I believe Ruby 2.6.2 and Rails 6 will run on cequel's master with no problems (at least for my use case).

You can safely delete this pull request, just keep in mind that at some point you'll need to replace `BigDecimal.new()` with `BigDecimal()` for Ruby 2.6.3+ compatibility.

PR #262, opened 2015-11-23
add support for datacenter configuration

muej:
Hi there,

I've added support for datacenter configuration. When used, the default load balancing policy Cassandra::LoadBalancing::Policies::DCAwareRoundRobin treats it as the local datacenter.

Kind regards
Jens

PR #416, opened 2019-09-16
Adding support for multiple DCs

ryakh:
Reviving https://github.com/cequel/cequel/pull/262

Adding support for multiple DC's with a `DCAwareRoundRobin` load-balancing policy.

### DIFF
--- a/lib/cequel/metal/keyspace.rb
+++ b/lib/cequel/metal/keyspace.rb
@@ -21,6 +21,8 @@ module Cequel
       attr_reader :hosts
       # @return Integer port to connect to Cassandra nodes on
       attr_reader :port
+      # @return [String] name of the datacenter
+      attr_reader :datacenter
       # @return [Symbol] the default consistency for queries in this keyspace
       # @since 1.1.0
       attr_writer :default_consistency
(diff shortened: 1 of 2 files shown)


spec/examples/metal/keyspace_spec.rb
```diff
@@ -86,6 +86,28 @@
     end
   end
 
+  describe "#configure" do
```
wflanagan, line 89: @ryakh This looks fine to me. I'm wondering if you can create a small test that pushes some data back and forth in both of these cases, so we all can be sure it works?

lib/cequel/metal/keyspace.rb
```diff
@@ -375,24 +391,28 @@ def extract_ssl_config(configuration)
...
         end
       end
-      
+
       def extract_cassandra_options(configuration)
         configuration[:cassandra_options]
       end
+
+      def extract_datacenter(configuration)
+        configuration.fetch(:datacenter, nil)
```
ilanusse: I'd prefer just `configuration[:datacenter]`

spec/examples/metal/keyspace_spec.rb
ilanusse: +1

stefannibrasil:
@awj @ryakh would it make sense to merge this branch to main? We are referencing to the latest commit SHA on the main convertkit app for a while.

